PR #1, opened 2022-11-30
Tsdk 42 generate mnemonics and main wallet key (scala)

mgrand-topl:
This is a partial pull request focused on setting up the directory structure, build process and pipeline. Publishing to Maven central will be in a subsequent PR.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Automatic end of line normalization. Normalize to LF on check in.
+* text=auto
+
+# Force LF endings for shell scripts.
+*.sh text eol=lf
+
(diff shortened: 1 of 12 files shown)


nandotorterolo (approved):
Looks good to me
It seems you are doing this stuff many times, and maybe I can help you.
If you want, we can create a Tople seed using giter8, when we can add code of conduct, git configs, scalafmt configs....

mgrand-topl:
> Looks good to me It seems you are doing this stuff many times, and maybe I can help you. If you want, we can create a Tople seed using giter8, when we can add code of conduct, git configs, scalafmt configs....

I agree that we should have a template repo. I want to wait until we have a library that we can build and publish; the

PR #3, opened 2022-12-19
Tsdk 42 generate mnemonics and main wallet key (scala) 3

mgrand-topl:
## Purpose
This contains the version of crypto with no dependencies on other Topl packages

## Approach
All references to the models package have been removed.

## Testing
Run existing unit tests

## Tickets
* TSDK-42
* TSDK-220

[TSDK-42]: https://topl.atlassian.net/browse/TSDK-42?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TSDK-220]: https://topl.atlassian.net/browse/TSDK-220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/maven_release.yml
```diff
@@ -0,0 +1,25 @@
+name: Maven Central Release
+on: [release, push]
```
mundacho, line 2: This should be only on `release` for the final version.

crypto/src/main/scala/co/topl/crypto/generation/KeyInitializer.scala
```diff
@@ -1,81 +1,77 @@
...
-import co.topl.models.utility.HasLength.instances._
-import co.topl.models.utility.Sized
-import scodec.bits.BitVector
+import scodec.bits.{BitVector, ByteVector}
 import simulacrum.typeclass
 
 import java.util.UUID
+import scala.annotation.unused
 
 @typeclass trait KeyInitializer[SK] {
```
mundacho, line 12: If you are not using SK anymore in the class you can maybe remove the type parameter here

mundacho (approved):

.github/workflows/maven_release.yml
mgrand-topl: I added push so that I could test the CI workflow. What is the reason to leave `release` out  until the final version?

crypto/src/main/scala/co/topl/crypto/hash/Blake2b.scala
```diff
@@ -1,11 +1,14 @@
...
-import co.topl.models.utility.{Lengths, Sized}
 import org.bouncycastle.crypto.digests.Blake2bDigest
+import scodec.bits.ByteVector
 
+/**
+ * Blake2b-* hashing scheme
+ *
+ * @tparam D the sized digest type with an implicit Digest implementation
+ */
 abstract class Blake2bHash[D: Digest] extends Hash[Blake2b, D] {
```
SeanCheatham, line 12: This `abstract class` may no longer be needed.  It was used by Dion, but I abandoned it for Tetra because of its usage of `synchronized`.  In Tetra, I instead either create a `new Blake2b256()` as needed, or I pull from a pool of instances (which requires some extra cats work so may not apply in this case).

crypto/src/main/scala/co/topl/crypto/hash/hash.scala
```diff
@@ -47,22 +53,50 @@ package object hash {
...
+   * The scala type for SHA2-* hash schemes
+   */
   type Sha
 
   import digest.implicits._
 
+  /**
+   * Use this to compute Blake2b-256 (32 byte) hashes.
+   */
   val blake2b256: Hash[Blake2b, Digest32] = new Blake2bHash[Digest32] {}
```
SeanCheatham, line 71: (Related to my comment above - this was only really used in Dion)

SeanCheatham (approved):

PR #10, opened 2023-01-25
TSDK-280 Refactor TransactionBuilder to follow algebra/interpreter pattern

DiademShoukralla:
## Purpose

The proof of concept design of TransactionBuilder was implemented as a mock and didn't allow the consumer to specify a context they want to run in. This POC was refactored to follow the algebra/interpreter paradigm used throughout our ecosystem.

## Approach

- Added `F[_]` to the TransactionBuilder trait (algebra)
- Replaced MockTransactionBuilder with TransactionBuilderInterpreter (interpreter)
    - TransactionBuilderInterpreter takes in a DataApi instead of hardcoding the usage of MockDataApi
    - Each exposed function returns a value wrapped in F[_]
    - Implementation details of most functions remain the same. The exception is `constructUnprovenTransaction`
- Minor update to Ed25519
    - exposed `seedLength` in `EllipticCurveSignatureScheme`
    - When deriving a Key Pair using `Ed25519`, Pad the seed so it's length matches `seedLength`. An error occurs if it is not the correct size

## Testing

- Related tests did not exist prior to these changes and were not added as a part of it.
- Sanity checked that the project compiles
- Sanity tested in a worksheet file that using the TransactionBuilderInterpreter works as expected. Specifically that `constructUnprovenTransaction` constructs a simple 1 input 1 output unproven transaction as expected.

## Tickets
* Closes TSDK-280: Topl/Brambl#168


mundacho (approved):
Looks good to me

PR #12, opened 2023-02-06
TSDK-253: Transaction Syntax Validation

DiademShoukralla:
## Purpose

Added Transaction Syntax Validation (and tests) + code cleanup/update due to PB updates.

## Approach

**TransactionSyntaxInterpreter:**

- All the validators are essentially the same (minor changes due to using different models) except `proofTypeValidation`. This was replaced with `attestationValidation` 
    - Note that this validation only makes sense if the Attestation type is Predicate. Also, since a Transaction may or may not be proven the list of responses (proofs) may be smaller than the list of challenges (propositions), or a proof may be of type Proof.Value.Empty. Both is okay since an unproven transaction is still valid on the client side. TransactionAuthorizationInterpreter should be used to determine if a transaction is properly proven.
- Note: Signature is slightly different than in bifrost

**Other changes:**
- removed an AttestationMalformed check from credentialler since its no longer a requirement that list of proofs matches the length of propositions (The way signable bytes is derived changed since). 
- Removed a directory (src) since it should not have been there
- Many changes throughout code due to PB validation updates. For example, removed `.some` and `Some(__)`.

## Testing

- Moved over (and regactored/updated) existing tests. These tests are mostly copy-pasted from Bifrost, however, updated to match the newer signature of TransactionSyntaxInterpreter (return Either, F[_] is applicative instead of Sync) and the updates to the implementation as described above.
- Ensured ContainsSignableSpec tests still pass (sanity due to lots of PB updates)

## Tickets
* Closes TSDK-253


brambl-sdk/src/test/scala/co/topl/brambl/MockHelpers.scala
```diff
@@ -10,33 +9,27 @@ import co.topl.brambl.common.ContainsSignable.instances._
...
+import co.topl.brambl.models.transaction._
 import co.topl.quivr.api.{Proposer, Prover}
 import com.google.protobuf.ByteString
 import quivr.models.{Int128, Proof, SignableBytes, SmallData}
 
 trait MockHelpers {
 
   val outDatum: Datum.UnspentOutput =
-    Datum.UnspentOutput(Event.UnspentTransactionOutput(SmallData(ByteString.copyFrom("metadata".getBytes)).some).some)
+    Datum.UnspentOutput(Event.UnspentTransactionOutput(SmallData(ByteString.copyFrom("metadata".getBytes))))
```
SeanCheatham, line 20: This could be added in a different PR/ticket, but @nandotorterolo started on implementing some scalacheck Generators for the protobuf models.  They're currently in the Bifrost repo, but they could be copied over here instead.  They would help with generating test data instead of hardcoding.

SeanCheatham (approved):

brambl-sdk/src/test/scala/co/topl/brambl/MockHelpers.scala
DiademShoukralla: Thanks for letting me know! I opened TSDK-307 to track this work.

mundacho (approved):
Looks good to me

PR #13, opened 2023-02-09
publish generators

nandotorterolo:
## Purpose
Publish Model Generators

## Approach
- update protobuf-specs
- update quivr
- create generators for **Brambl**   and Quivr
- Next Step: same task on Quivr
- Next Step: remove Quivr generator on Brmbl, and use Quivr published

## TODO, 
-1 question about if we do want to publish both projects

## Testing
publishLocal

image

## Tickets
* TODO

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -48,7 +48,7 @@ lazy val publishSettings = Seq(
   homepage := Some(url("https://github.com/Topl/BramblSc")),
   sonatypeCredentialHost := "s01.oss.sonatype.org",
   sonatypeRepository := "https://s01.oss.sonatype.org/service/local",
-  Test / publishArtifact := true, // TODO Ask if we want to publish Test Sources of Crypto and Brambl, or only Bramble
+  Test / publishArtifact := false,
   pomIncludeRepository := { _ => false },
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -48,7 +48,7 @@ lazy val publishSettings = Seq(
   homepage := Some(url("https://github.com/Topl/BramblSc")),
   sonatypeCredentialHost := "s01.oss.sonatype.org",
   sonatypeRepository := "https://s01.oss.sonatype.org/service/local",
-  Test / publishArtifact := false,
+  Test / publishArtifact := true, // TODO Ask if we want to publish Test Sources of Crypto and Brambl, or only Bramble
```
SeanCheatham, line 51: I think it may be better to keep this defaulted to `false` here in `publishSettings`, and then override it to true _just_ in the `brambl-sdk`

mundacho (approved):

DiademShoukralla (approved):

build.sbt
nandotorterolo: thanks for reviewing it, changed it.

PR #14, opened 2023-02-10
TSDK-279 CredentiallerInterpreter

DiademShoukralla:
## Purpose

The proof of concept design of Credentialler was implemented as a mock and didn't allow the consumer to specify a context they want to run in. This POC was refactored to follow the algebra/interpreter paradigm used throughout our ecosystem.

## Approach

- Added `F[_]` to the Credentialler trait (algebra)
- Modified signatures related to `prove` since there's no reason to have a error during prove; if a proposition is unprovable, the proof will be Proof.Value.Empty
- Replaced MockCredentialler with CredentiallerInterpreter (interpreter)
    - CredentiallerInterpreter takes in a DataApi instead of hardcoding the usage of MockDataApi
    - Each function returns a value wrapped in F[_]
    - Implementation details of most functions remain the same. The exception is `validate` also validates transaction syntax in addition to authorization. 
- Context takes in F[_] instead of hardcoding Id
- Cleaned up and moved MockDataApi to the test directory. Removed mock data that it no longer uses and moved the remaining mock data to MockHelpers. MockDataApi will use the mock data from MockHelpers
    - Due to changes in MockHelpers, I updated the existing tests that were affected
- Implemented bare minimum unit tests for CredentiallerInterpreter for sanity
- Code cleanup

## Testing

- Implemented CredentiallerInterpreterSpec with basic unit tests. These tests do not test all possible use cases for Credentialler since they have not yet been settled; instead they serves as sanity tests
- Ensured all existing tests still pass

## Tickets
* Closes TSDK-279: Topl/Brambl#169


mundacho (approved):
LGTM! 👍

ignaciofernandezsoto (approved):
LGTM!!

PR #22, opened 2023-02-21
TSDK-328 Fix flaky unit test

DiademShoukralla:
## Purpose

A test unit test was not consistent. When running the test via Intellij IDE, it was consistently passing. However, the test is run via SBT (including Github actions), the result was inconsistent (sometimes passing, sometimes failing). 

Debugging revealed that the issue is caused by Ed25519Signature.createKeyPair (Ed25519.instance.deriveKeyPairFromSeed) not being deterministic when run in SBT. Specifically, the Verification Key returned was not the one expected. It came up today in EOD check-in that Ed25519.instance was not thread-safe.

## Approach

- Updated the unit test to be more verbose
- Replaced the usage of `Ed25519.instance` with `new Ed25519`

## Testing

- Updated the flaky test to be more verbose
- Manually ran the test suite using **SBT** 15 times consecutively to ensure the test _consistently_ passes

## Tickets
* TSDK-328: Topl/Brambl#218


nandotorterolo:
Thanks for the Fix

nandotorterolo (approved):

PR #23, opened 2023-02-22
update-scala-steward-failedJob

nandotorterolo:
## Purpose
This PR, is the same that https://github.com/Topl/BramblSc/pull/19

But it is aligned with main, which includes the last change with a Flaky test fix. #22 
## Note
if we merge this, close failed PR https://github.com/Topl/BramblSc/pull/20

## Testing
preparePR
## Tickets
*

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = "3.5.3"
+version = "3.5.9"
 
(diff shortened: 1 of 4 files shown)


DiademShoukralla (approved):
Thanks for this! Looks good!

PR #27, opened 2023-03-10
Box Values, Remove KnownIdentifier, Generators

SeanCheatham:
## Purpose
- Update protobuf-specs
- Removes usage of KnownIdentifier
- Adds a few more transaction-related generators
- Adds serialization for new Box Value types
## Approach
- Use new protobuf-specs version
- Replace KnownIdentifier with TransactionOutputAddress
## Testing
- preparePR
- Used in Bifrost
## Tickets
- BN-713

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
@@ -1,23 +1,29 @@
 package co.topl.brambl.builders
 
 import cats._
+import cats.data.EitherT
+import cats.data.Validated
 import cats.implicits._
-import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxEitherObject, catsSyntaxOptionId, toFlatMapOps}
+import cats.implicits.catsSyntaxApplicativeId
+import cats.implicits.toFlatMapOps
 import co.topl.brambl.common.ContainsEvidence.Ops
 import co.topl.brambl.common.ContainsImmutable.instances._
 import co.topl.brambl.dataApi.DataApi
-import co.topl.brambl.models.{Address, Datum, Event, Identifier, KnownIdentifier}
+import co.topl.brambl.models.LockAddress
+import co.topl.brambl.models.TransactionOutputAddress
 import co.topl.brambl.models.box.Lock
-import co.topl.brambl.models.builders.{InputBuildRequest, OutputBuildRequest}
-import co.topl.brambl.models.transaction.{
-  Attestation,
-  IoTransaction,
-  Schedule,
-  SpentTransactionOutput,
-  UnspentTransactionOutput
-}
-import quivr.models.{Proof, SmallData}
-import cats.data.{EitherT, Validated}
+import co.topl.brambl.models.builders.InputBuildRequest
+import co.topl.brambl.models.builders.OutputBuildRequest
+import co.topl.brambl.models.transaction.Attestation
+import co.topl.brambl.models.transaction.IoTransaction
+import co.topl.brambl.models.transaction.Schedule
+import co.topl.brambl.models.transaction.SpentTransactionOutput
+import co.topl.brambl.models.transaction.UnspentTransactionOutput
+import co.topl.brambl.models.Datum
+import co.topl.brambl.models.Event
+import co.topl.brambl.models.Identifier
+import quivr.models.Proof
+import quivr.models.SmallData
 
 object TransactionBuilderInterpreter {
 
@@ -26,20 +32,20 @@ object TransactionBuilderInterpreter {
     override def constructUnprovenInput(
       data: InputBuildRequest
     ): F[Either[BuilderError.InputBuilderError, SpentTransactionOutput]] = {
-      val box = dataApi.getBoxByKnownIdentifier(data.id)
+      val box = dataApi.getBoxByKnownIdentifier(data.address)
       val attestation = box.map(_.lock).map(constructUnprovenAttestation)
       val value = box.map(_.value)
       val datum = Datum.SpentOutput(Event.SpentTransactionOutput(data.metadata.getOrElse(SmallData())))
       (attestation, value) match {
         case (Some(Right(att)), Some(boxVal)) =>
-          SpentTransactionOutput(data.id, att, boxVal, datum)
+          SpentTransactionOutput(data.address, att, boxVal)
             .asRight[BuilderError.InputBuilderError]
             .pure[F]
         case (Some(Left(err)), _) => err.asLeft[SpentTransactionOutput].pure[F]
         case _ =>
           BuilderError
             .InputBuilderError(
-              s"Could not construct input. Id=${data.id}, Attestation=${attestation}, Value=${value}"
+              s"Could not construct input. Id=${data.address}, Attestation=${attestation}, Value=${value}"
             )
             .asLeft[SpentTransactionOutput]
             .pure[F]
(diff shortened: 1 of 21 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/common/ContainsImmutable.scala
```diff
@@ -95,33 +96,35 @@ object ContainsImmutable {
 
     implicit val iotxScheduleImmutable: ContainsImmutable[Schedule] = (schedule: Schedule) =>
       schedule.min.immutable ++
-      schedule.max.immutable
+      schedule.max.immutable ++
+      schedule.timestamp.immutable
```
SeanCheatham, line 100: I think it was discussed previously to _not_ include the timestamp in the signable/identifiable/immutable bytes, but I'd prefer to include it to help create "unique" transactions

brambl-sdk/src/test/scala/co/topl/brambl/validation/TransactionSyntaxInterpreterSpec.scala
```diff
@@ -177,49 +165,4 @@ class TransactionSyntaxInterpreterSpec extends munit.FunSuite with MockHelpers {
       .exists(_.toList.contains(TransactionSyntaxError.InvalidDataLength))
     assertEquals(result, true)
   }
-
```
SeanCheatham, line 180: This test would be hard to re-create since there's only space for at most 64-bytes in SmallData now

DiademShoukralla (approved):
LGTM

brambl-sdk/src/main/scala/co/topl/brambl/common/ContainsImmutable.scala
scasplte2: we should not include timestamp

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
```diff
@@ -91,8 +95,6 @@ object TransactionBuilderInterpreter {
             Event.IoTransaction(
               // TODO: Replace min and max with slot numbers
               schedule.getOrElse(Schedule(0, 2147483647, System.currentTimeMillis)),
```
scasplte2, line 85: Wouldn't this maximum limit us to approximately 10 years of slots at a one second slot duration?

brambl-sdk/src/main/scala/co/topl/brambl/syntax/LockSyntax.scala
```diff
@@ -0,0 +1,40 @@
...
+  implicit def lockAsLockSyntaxOps(lock: Lock): LockSyntaxOps =
+    new LockSyntaxOps(lock)
+
+  implicit def predicateLockAsLockSyntaxOps(lock: Lock.Predicate): PredicateLockSyntaxOps =
+    new PredicateLockSyntaxOps(lock)
+}
+
+class LockSyntaxOps(val lock: Lock) extends AnyVal {
+
+  def address(network: Int, ledger: Int): LockAddress =
```
scasplte2, line 22: we need to determine if Lock addresses would be different for a the same proposition in the three different constructions

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionAuthorizationInterpreter.scala
```diff
@@ -118,7 +118,7 @@ object TransactionAuthorizationInterpreter {
         context:      DynamicContext[F, String, Datum]
       )(implicit verifier: Verifier[F, Datum]): F[Either[TransactionAuthorizationError, Boolean]] =
         if (threshold === 0) true.asRight[TransactionAuthorizationError].pure[F]
-        else if (threshold >= propositions.size)
+        else if (threshold > propositions.size)
```
scasplte2, line 130: a threshold is at or above, no?

scasplte2 (changes requested):
I would like to have a discussion about the bit ordering definitions. It is critical that we have an understanding across the team about this

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
SeanCheatham: Yes, but I'll let that be taken care of in a different PR since there's already a TODO

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionAuthorizationInterpreter.scala
SeanCheatham: The way it was defined before, if the threshold=1 and 1 proposition was provided, a `Left` would be returned which fails simple Transaction cases.

brambl-sdk/src/main/scala/co/topl/brambl/common/ContainsImmutable.scala
SeanCheatham: Updated PR

brambl-sdk/src/main/scala/co/topl/brambl/syntax/LockSyntax.scala
SeanCheatham: I assume the three (five) different Lock constructions would result in different addresses since they contain different data.

scasplte2: but if they are based on the same proposition(s), does it help discovery / persistence? To do this, I think you have to go `Predicate -> Image -> Commitment` and take the signable bytes of the commitment that is hashed. I agree the easier method is to define signable bytes for each of the different Lock types though.

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionAuthorizationInterpreter.scala
scasplte2: hmm... should we have a separate case for `threshold === propositions.size`? I thought this code was copied directly from your prior implementation of Propositions and Proofs from the proof of concept

SeanCheatham: I believe the logic is sufficiently handled by the `else` case.  I'm not sure where this implementation came from, but it doesn't resemble Bifrost's [original](https://github.com/Topl/Bifrost/blob/dev/ledger/src/main/scala/co/topl/ledger/interpreters/TransactionAuthorizationValidation.scala#L161) threshold verifier.

brambl-sdk/src/main/scala/co/topl/brambl/syntax/LockSyntax.scala
SeanCheatham: I'm not sure - I'm still working on grasping all the different lock constructions anyway.  My only concern right now is to get a very basic lock working so that Bifrost can **start** working with the new models.  If changes need to be made, I'm sure we can make them.  If you need me remove this code here, I can; I'll just add in something separate in Bifrost in the interim.  I only added these syntax helpers to avoid re-writing the same code multiple times in Bifrost.

brambl-sdk/src/main/scala/co/topl/brambl/common/ContainsImmutable.scala
```diff
@@ -1,9 +1,12 @@
 package co.topl.brambl.common
```
SeanCheatham, line 1: @scasplte2 
> I would like to have a discussion about the bit ordering definitions. It is critical that we have an understanding across the team about this

Could you clarify the scope of the discussion you'd like to have?  Are there specific changes you would like made, and do those changes need to happen in this PR?

scasplte2: I don't recall, so it must not have been important enough for me to remember

PR #29, opened 2023-03-17
TSDK-352 Added ExtendedEd25519 to Ecosystem Crypto

DiademShoukralla:
## Purpose

ExtendedEd25519 is needed for the HD Wallet (child derivation). This did not exist in the current Ecosystem crypto library but did in a previous commit on bifrost. The purpose of this PR is to move the implementation over (along with the tests). 

## Approach

- Refactor of EllipticCurveSignatureScheme such that both Ed25519 and ExtendedEd25519 can extend it. This involved changes to Ed25519 as well
- Removal of all ByteVector usage. Replaced with Array[Byte] and specific case classes for Secret/Public key for both implementations. Removal of ByteVector (and new models) also required updates to Entropy, EntropyToSeed, Bip32Index, Hex, and brambl-sdk
- Added ExtendedEd25519 implementation based on this prior implementation: https://github.com/Topl/Bifrost/blob/20735a492f6e6eb0681c5ecc6c84a925ea91ff24/crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
- Added ExtendedEd25519 tests based on this prior implementation: https://github.com/Topl/Bifrost/blob/20735a492f6e6eb0681c5ecc6c84a925ea91ff24/crypto/src/test/scala/co/topl/crypto/signing/ExtendedEd25519Spec.scala
- Some ExtendedEd25519 required KeyInitializer to have the ExtendedEd25519 implementation so that was added in this PR based on: https://github.com/Topl/Bifrost/blob/20735a492f6e6eb0681c5ecc6c84a925ea91ff24/crypto/src/main/scala/co/topl/crypto/generation/KeyInitializer.scala (along with its tests). This covers partial work for TSDK-354

## Testing

- Ensure all existing tests pass (tests in crypto and brambl-sdk)
- Ensure changes does not break brambl-sdk (fixed changes)
- Added ExtendedEd2559 tests for Signing and KeyInitializer

## Tickets
* Closes TSDK-352
* Partial work for TSDK-354


crypto/src/main/scala/co/topl/crypto/signing/EllipticCurveSignatureScheme.scala
scasplte2, line 1: Nicely done 🔥

crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
```diff
@@ -0,0 +1,358 @@
...
+   * Sign a given message with a given signing key.
+   *
+   * @note Precondition: the private key must be a valid ExtednedEd25519 secret key
+   * @note Postcondition: the signature must be a valid ExtendedEd25519 signature
+   *
+   * @param privateKey The private signing key
+   * @param message    a message that the the signature will be generated for
+   * @return the signature
+   */
+  override def sign(privateKey: SecretKey, message: Array[Byte]): Array[Byte] = {
```
scasplte2, line 28: I thought this could be replaced via Bouncy Castle. Is the `eddsa` down below in `impl` still the recreated version that Aaron made?

crypto/src/main/scala/co/topl/crypto/signing/signing.scala
scasplte2, line 1: would it help anything in Quivr to pull these models down to `common` models at some point?

scasplte2 (approved):
💯

crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
DiademShoukralla: I tracked the original commit where the [implementation was added](https://github.com/Topl/Bifrost/commit/c7cbe46049ea419672678d36e2bf25a39597f49e#diff-0f1efeab6a457b8f0c8c39a809c9ec5e1093b2e3115551176f3f52b95cd730c4) and I believe it is the one you are referring to (that Aaron made).
I'm looking into if it is a straight forward change to replace it with BouncyCastle

DiademShoukralla: @scasplte2 The replacement is straightforward for Ed25519 but not ExtendedEd25519 as the latter requires access to functions in the [BouncyCastle implementation](https://javadoc.io/static/org.bouncycastle/bcprov-jdk15on/1.64/org/bouncycastle/math/ec/rfc8032/Ed25519.html) that are inaccessible (marked as `private static`). 

For instance (where `impl = rfc8032.Ed25519`) , 
image

image

I created a new ticket, TSDK-406, to track the work that will be needed to do this replacement.

hwbehrens (approved):
Looks good! Thanks for taking care of good ol' ExtendedEd25519.

SeanCheatham (approved):

PR #30, opened 2023-03-21
TSDK-354 Remove Scodec from KeyInitializer

DiademShoukralla:
## Purpose

TSDK-354 requires KeyInitializer to support ExtendedEd25519. Most of that work was done in https://github.com/Topl/BramblSc/pull/29  . The remaining work to complete KetInitializer is to remove the usage of Scodec

## Approach

- Removed Scodec import from KeyInitializer
- Moved Hex.scala from /test/ directory into main code within new package "utils"
- Replaced BitVector.fromHex(base16String) with Hex.hexStringToStrictBytes(base16String)

## Testing
Ensured all existing tests pass

## Tickets
* Closes TSDK-354


scasplte2 (approved):
🧱 ➡️ ⛰️

hwbehrens (approved):
Looks good! Nice, small PR :)

crypto/src/main/scala/co/topl/crypto/generation/KeyInitializer.scala
```diff
@@ -51,16 +48,16 @@ trait KeyInitializer[SK <: SigningKey] {
...
-      .fromOption(BitVector.fromBase58(base58String), InitializationFailures.InvalidBase58String)
-      .map(bits => fromBytes(bits.toByteArray))
-
+  /**
+   * Create a secret key from a Base16 Hex string.
+   *
+   * @param base16String The Hex string which encodes a Secret key
+   * @return The created secret key if base16String was a valid Base16 Hex String. Else an InitializationFailure error.
+   */
   def fromBase16String(base16String: String): Either[InitializationFailure, SK] =
```
SeanCheatham, line 57: I'm curious if we need to keep this `fromBase16String` method, or if we should only use the `fromBytes` and expect the base16 -> Array[Byte] to happen externally?

DiademShoukralla: I have no problem removing it since the base16 -> Array[Byte] can be done externally pretty easily with utils.Hex. If we want to remove it, it would be good to do it now since the only place this function is currently called is within the tests

SeanCheatham: My main motivation to remove it was to keep the utils.Hex down in the unit test code.  I don't have a huge preference either way though.

SeanCheatham (approved):

PR #31, opened 2023-03-27
TSDK-356 Add AES and SCrypt

DiademShoukralla:
## Purpose

For Encrypting a Keyfile, we need to have an implementation of a Cipher and a KDF (see image below). We will use AES as the cipher, and SCrypt as the KDF for the initial implementation.

image

## Approach

- Added `cipher` package with `Cipher` trait and `kdf` package with `Kdf` trait in a new `encryption` package. `Cipher` has `encrypt` and `decrypt` while `Kdf` has `deriveKey`. Since different implementations of ciphers and kdfs will have their own parameters, the `Param` trait in each is empty and will be extended by the specific implementations.
- Added `Aes` in the `encryption.cipher` package. 
- Added SCrypt in the `encryption.kdf` package. 
- The implementation details of both were informed by an older implementation of generating a KeyFile. See https://github.com/Topl/Bifrost/blob/42fcf1e7d4d68d1861b76720fc1007dbacfd8eae//brambl/src/main/scala/co/topl/credential/KeyFile.scala 
- These changes do not introduce any breaking changes to clients since it simply adds additional packages. No existing code was altered.

## Testing

I added unit tests for both Aes and Scrypt. 

## Tickets
* Closes TSDK-356

### DIFF
--- a/crypto/src/main/scala/co/topl/crypto/encryption/cipher/Aes.scala
+++ b/crypto/src/main/scala/co/topl/crypto/encryption/cipher/Aes.scala
@@ -1,19 +1,57 @@
 package co.topl.crypto.encryption.cipher
 
+import org.bouncycastle.crypto.engines.AESEngine
+
 /**
  * AES encryption.
  * Aes is a symmetric block cipher that can encrypt and decrypt data using the same key.
  * @see [[https://en.wikipedia.org/wiki/Advanced_Encryption_Standard]]
  */
 class Aes extends Cipher[Aes.AesParams]{
-  override def encrypt(plainText: Array[Byte], params: Aes.AesParams): Array[Byte] = ???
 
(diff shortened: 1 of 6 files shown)


mundacho (approved):
Looks good to me, I just noticed that you used a very direct approach and no F[_]s and other functional stuff. Any particular reason?

DiademShoukralla:
@mundacho Good catch! The reason I didn't use F[ _ ]s or the interpreter/algebra pattern is because that paradigm isn't yet used in any other part of the Ecosystem Crypto library and I wanted it to be as consistent as possible. However, I am not opposed to adding that in for Cipher and Kdf in this PR. 
If I do, do we also want to plan to use F[ _ ]s and the interpreter/algebra paradigm in the other parts (signing, hash, generation, and accumulators package) of the library as well? 

@hwbehrens Do you have any opinion on this?

PR #37, opened 2023-04-05
TSDK-357 Add Support for VaultStore (formally KeyFile)

DiademShoukralla:
## Purpose

Sensitive information (such as the Topl main keys) will need to be stored in an encrypted manner. When we encrypt secret information (using a KDF and Cipher) we will use a VaultStore to keep all the information needed to decrypt it at a later time. The idea of the VaultStore (formally KeyFile) is not a novel idea to the ecosystem (see previous [KeyFile discovery](https://topl.atlassian.net/wiki/external/429391942/MGQyN2M5MGYzNDc3NGU1NTliMmVjYTgyMjlhYjFjNGI))

## Approach

- cipher and kdf defiinitions were added in a previous PR. In this PR I refactored them to use F[_] (more functional approach). Aes and SCrypt now have interpreters.
- Added algebra/interpreter for a Mac (needed for VaultStore). A mac is used to validate the integrity of the decrypted secret
- Using a Kdf, Cipher, mac, and cipher text, created a VaultStore. The VaultStore can decrypt the cipher text it contains using the other parameters
- Added Json Codecs for Aes, SCrypt, Cipher, Kdf, and VaultStore
- Added and updated tests

## Testing
- Added tests for Mac, Vaultstore, and Codecs

## Tickets
* Closes TSDK-357


crypto/src/test/scala/co/topl/crypto/encryption/CodecSpec.scala
```diff
@@ -0,0 +1,268 @@
...
+    val testParams = Helpers.ExpectedAesParams.json.as[Aes.AesParams]
+    testParams.isRight shouldBe true
+    testParams shouldBe (Right(Helpers.ExpectedAesParams.value))
+    // Encode test
+    val testJson = Helpers.ExpectedAesParams.value.asJson
+    testJson.isNull shouldBe false
+    testJson shouldBe Helpers.ExpectedAesParams.json
+
+    // Decode then Encode test
+    val encodedFromDecoded = testParams.map(_.asJson).toOption
```
SeanCheatham, line 30: Since you're using Scalatest, you can mixin the trait `OptionValues` to this test class to give you a `.value` helper method on `Option`s.  There's also `EitherValues` for working with Eithers.

But if there's ever a plan to migrate this to use mUnit, it would be better to keep it as-is.

crypto/src/main/scala/co/topl/crypto/encryption/Mac.scala
```diff
@@ -0,0 +1,54 @@
...
+   */
+  def validateMac[F[_]: Applicative](expectedMac: Array[Byte]): F[Boolean]
+
+  /**
+   * Validate the MAC.
+   *
+   * @param expectedMac the expected MAC
+   * @return true if the MAC is valid, false otherwise
+   */
+  def validateMac[F[_]: Applicative](expectedMac: Mac): F[Boolean]
```
nandotorterolo, line 30: I am not clear with the method, because if you have Mac, means that it was created previously with the method calculatedMac, which maybe I am confused gut It will always return true, dismiss my comment If I am wrong

crypto/src/main/scala/co/topl/crypto/encryption/Mac.scala
```diff
@@ -0,0 +1,54 @@
...
+  // The MAC
+  val value: Array[Byte]
+
+  /**
+   * Validate the MAC.
+   *
+   * @param expectedMac the expected MAC value
+   * @return true if the MAC is valid, false otherwise
+   */
+  def validateMac[F[_]: Applicative](expectedMac: Array[Byte]): F[Boolean]
```
nandotorterolo, line 22: maybe the parameter could be named:  `expectedMacValue`
both methods use the name param name expectedMac, with  different types: Array[Byte], Mac
See my comment below, I think the other method is not necessary.

nandotorterolo (approved):

crypto/src/main/scala/co/topl/crypto/encryption/Mac.scala
DiademShoukralla: This method is meant to validate the mac against a provided expected mac (that may be different). 

In practice, the expected mac should be the mac that was generated at the time the VaultStore was generated (this value is stored in the VaultStore). At a later time when, the VaultStore is being decrypted, we calculate a new mac and compare with the expected mac to ensure the integrity of the decrypted value. 

I will update the scaladoc to be more clear about it's purpose and use

crypto/src/test/scala/co/topl/crypto/encryption/CodecSpec.scala
DiademShoukralla: Thanks for the suggestion!

PR #39, opened 2023-04-12
Wallet fix

DiademShoukralla:
## Purpose

Update WalletApi.createNewWallet based on some recent feedback

## Approach

- When returning the mnemonic, return IndexedSeq[String] instead of a single String
- Use `for` comprehension in `createNewWallet`
- Update return type such that both types of errors will return (previously if a DataApi error occurs, the caller will not know)
- Update return type such that the VaultStore is also returned as well (in addition to mnemonic)

## Testing

- Ensured all existing tests will pass
- ran `checkPR`

## Tickets
* N/A


mundacho (approved):
Great!

PR #40, opened 2023-04-12
TSDK-413 Removed all remaining usages of Scodec

DiademShoukralla:
## Purpose

We no longer want to depend on Scodec

## Approach

Removed any remaining usages of Scodec and replaced with appropriate data types. Removed Scodec from dependencies

## Testing

- Any tests that depended on scodec were updated as well
- Ran `checkPR`, ensured it passed

## Tickets
* Part of TSDK-413 => Updates to Bifrost are still required


mundacho (approved):
Looks good to me!

PR #41, opened 2023-04-13
Try to improve the design

mundacho:
## Purpose

There was a complication in the implementation of the generic `createAndSaveNewWallet`.

## Approach

I created the function using some type parameters so that it supports Monad operations

## Testing

The tests were fixed.

## Tickets

* No tickets


DiademShoukralla (approved):
Looks great, thanks!

PR #43, opened 2023-04-14
Made WalletApiErrors extend RuntimeError

DiademShoukralla:
## Purpose

It is difficult to combine different operations because all the error types are different. A solution would be to make all the errors extend Throwable/RuntimeError.

## Approach

DataApiError already extends RuntimeError. This PR changes WalletApiError to also extend RuntimeError

## Testing

Ran `checkPR` and ensured was successful 

## Tickets
* n/a


mundacho (approved):
👍

PR #45, opened 2023-04-19
TSDK-437 Finish up recent changes to WalletApi

DiademShoukralla:
## Purpose

Part 2 of the recent changes to WalletApi

## Approach

- Removed `/routine/` package because I didn't feel like it was necessary (it essentially contained wrappers around functions in the crypto library; they can be used directly instead)
- Changes `Ed25519SignatureInterpreter` to `ExtendedEd25519SignatureInterpreter` since the extended version is what we should be using due to HD Wallet
- Removed `getKeyPair` from `DataApi` since a corresponding function now exists in `WalletApi`
- Updated `CredentiallerInterpreter` to use the `WalletApi` for proving Signature propositions (utilizing `deriveChildKeyPair`). Added a mainKey (`KeyPair`) parameter to the initialization of the `CredentiallerInterpreter`; to be able to prove transactions you need access to your main key (essentially be "logged on"; have wallet loaded).
- Moved the implicit PB <> Crypto conversions to the WalletApi static object and removed private modifier so it can be used elsewhere in the codebase
- Updated `MockDataApi` and `MockHelpers` to align with the changes 
- Added an additional test to the `CredentiallerInterpreterSpec` to test initializing the `CredentiallerInterpreter` with a different main key. This test revealed a bug within `CredentiallerInterpreter` where if there were no transaction syntax errors (but there were authorization errors) the transaction would still be considered valid. Modified the `validate` function within `CredentiallerInterpreter` to fix
- Added some unit tests for the new functions within the `WalletApi`

## Testing

- Added new tests to WalletApiSpec (to reflect the newly added functions)
- Added a new test to CredentiallerInterpreterSpec
- ran `checkPR` and ensured was successful
- Tested that these changes do not break bifrost (updated bifrost's BramblSc dependency and ran `checkPR`)

## Tickets
* Closes TSDK-437


mundacho (approved):
LGTM!

PR #46, opened 2023-04-20
TSDK-419 Update Wallet + Update Wallet's Password

DiademShoukralla:
## Purpose
 
Made changes to WalletApi and DataApi to support the use case of updating a wallet's password

## Approach

- added `updateMainKeyVaultStore` to DataApi
- Changed WalletApi's private `buildMainKeyVaultStore` to be exposed in the Algebra. 
- Added 2 additional functions to WalletApi's algebra; `updateWallet` and `updateWalletPassword`. The latter is implemented in the trait. The former is implemented in `make` using the new `DataApi.updateMainKeyVaultStore`
- Updated MockDataApi for tests to reflect the change in DataApi. I also mocked a way to trigger a save failure in `saveMainKeyVaultStore`
- Updated MockHelpers to re-instantiate `new ExtendedEd25519` each time its used (to fix flaky tests)
- Added unit tests for all the new functions + added an `afterEach` cleanup function to reset the persisted data in MockDataApi

## Testing

- Added unit tests
- ensured all existing tests passed
- ran `checkPR` and ensured successful

## Tickets
* Closes TSDK-419

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
@@ -79,6 +79,16 @@ trait DataApi[F[_]] {
    * @return The VaultStore for the Topl Main Secret Key if it exists. If retrieving fails due to an underlying cause, return a DataApiException
    */
   def getMainKeyVaultStore(name: String): F[Either[DataApiException, VaultStore[F]]]
+
+  /**
+   * Update a persisted VaultStore for the Topl Main Secret Key.
+   *
+   * @param name              The name identifier of the VaultStore to update. This is used to manage multiple wallet identities.
+   *                          Most commonly, only one wallet identity will be used. It is the responsibility of the dApp
+   *                          to manage the names of the wallet identities if multiple will be used.
+   * @return nothing if successful. If the update fails due to an underlying cause (for ex does not exist), return a DataApiException
+   */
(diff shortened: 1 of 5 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
```diff
@@ -79,6 +79,16 @@ trait DataApi[F[_]] {
    * @return The VaultStore for the Topl Main Secret Key if it exists. If retrieving fails due to an underlying cause, return a DataApiException
    */
   def getMainKeyVaultStore(name: String): F[Either[DataApiException, VaultStore[F]]]
+
+  /**
+   * Update a persisted VaultStore for the Topl Main Secret Key.
+   *
+   * @param name              The name identifier of the VaultStore to update. This is used to manage multiple wallet identities.
+   *                          Most commonly, only one wallet identity will be used. It is the responsibility of the dApp
+   *                          to manage the names of the wallet identities if multiple will be used.
```
mundacho, line 88: is the name a file name or can it be something else?

mundacho (approved):
LGTM!

brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
DiademShoukralla: It can be a file name or it can be something else. It would be up to the DataApi implementation to decide how it will be used

PR #51, opened 2023-04-27
Implement TransactionCostCalculator

SeanCheatham:
## Purpose
- Some transactions are “harder” to validate than others.  In particular, transactions which spend a lot of inputs, produce a lot of outputs, or use complicated contracts may take a lot more CPU/memory/bandwidth to evaluate than a simple transfer.
- A function should exist which estimates how hard it would be to include a given transaction in the blockchain.
## Approach
- Create TransactionCostCalculator and TransactionCostCalculatorInterpreter
- Estimate cost of a Transaction based on its immutable bytes, difficulty of input proofs, and number of outputs
- Create initial set of cost estimates for proof types
## Testing
- New basic unit tests
## Tickets
- #BN-900


DiademShoukralla (approved):

mundacho (approved):
LGTM!

PR #52, opened 2023-04-27
TSDK-467 Add encoding functions

mundacho:
## Purpose

The goal is to provide encoding functions for the address codecs.

## Approach

We implemented the Base58, and Base58Check algorithm.

## Testing

Unit tests were added.

## Tickets

* TSDK-467


DiademShoukralla (approved):
Looks great!

PR #53, opened 2023-04-28
TSDK-460: Update how Credentialler retrieves information

DiademShoukralla:
## Purpose

Previously the Credentialler was under the assumption that indices (x/y/z) were associated to a UTXO. We know now that this is not the case as indices are associated to a Key Pair which can map to one or more SignaturePropositions and a UTXO's lock may contain many SignaturePropositions.

## Approach

This PR contains changes to Credentialler, TransactionBuilder, and DataApi. At a high level, instead of the Credentialler retrieving the indices for a utxo, the Credentialler will retrieve the individual secrets needed for each proposition.

Specifically: 
- A SignatureProposition will map to Indices in the DataApi
- A DigestProposition will map to a Preimage in the DataApi

DataApi was updated to include functions to support this:
- Remvoed `getIndicesByTxoAddress`  since a UTXO/STXO is not directly associated with indices
- Changed return types to be F[Either[XX, YY]]] instead of Option[YY] to utilize the F context as well as to be consistent with  the return types of the other DataApi functions
- Added functions `getPreimage` and `getIndices` to support the functionality needed in Credentialler

Additionally, TransactionBuilder.constructUnprovenInput was modified as well to conform with the updated DataApi signatures.

## Testing

- Added a new test for TransactionBuilder to ensure failures are captured properly and updated a test for Credentialler. Lots of changes to MockHelpers and MockDataApi were needed
- Ran `checkPR` and ensure was successful

## Tickets
* Closes TSDK-460

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
@@ -29,7 +29,7 @@ object TransactionBuilderInterpreter {
       (
         for {
           utxo <- EitherT(dataApi.getUtxoByTxoAddress(data.address))
-          lock <- EitherT(dataApi.getLock(utxo.address.id))
+          lock <- EitherT(dataApi.getLockByLockAddress(utxo.address))
         } yield (constructUnprovenAttestation(lock), utxo.value)
       ).value map {
         case Left(err) =>
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
@@ -33,7 +33,7 @@ object TransactionBuilderInterpreter {
         } yield (constructUnprovenAttestation(lock), utxo.value)
       ).value map {
         case Left(err) =>
-          BuilderError.InputBuilderError("Unable to construct Attestation", err).asLeft[SpentTransactionOutput]
+          BuilderError.InputBuilderError("Unable to construct proven input", err).asLeft[SpentTransactionOutput]
         case Right((Left(err), _)) => err.asLeft[SpentTransactionOutput]
         case Right((Right(att), value)) =>
           SpentTransactionOutput(data.address, att, value).asRight[BuilderError.InputBuilderError]
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
@@ -25,25 +25,19 @@ object TransactionBuilderInterpreter {
 
     override def constructUnprovenInput(
       data: InputBuildRequest
-    ): F[Either[BuilderError.InputBuilderError, SpentTransactionOutput]] = {
-      val utxo = dataApi.getUtxoByTxoAddress(data.address)
-      val attestation = utxo.map(_.address).flatMap(dataApi.getLockByLockAddress).map(constructUnprovenAttestation)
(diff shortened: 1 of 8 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
```diff
@@ -18,46 +18,43 @@ import quivr.models.Preimage
...
    * Return the UTXO targeted by a TransactionOutputAddress.
    *
    * A TransactionOutputAddress identifies an output (UTXO) of an existing transaction on the chain.
    *
    * @param address The TransactionOutputAddress of the UTXO to retrieve
-   * @return The UTXO targeted by the given address, if it exists. Else None
+   * @return The UTXO targeted by the given address, if it exists. Else a DataApiException
    */
-  def getUtxoByTxoAddress(address: TransactionOutputAddress): Option[UnspentTransactionOutput]
+  def getUtxoByTxoAddress(address: TransactionOutputAddress): F[Either[DataApiException, UnspentTransactionOutput]]
```
mundacho, line 29: This might need to return a list of UTXOs

brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
```diff
@@ -18,46 +18,43 @@ import quivr.models.Preimage
...
+   * verification and must be signed with the corresponding signing key. The verification and signing key pair is
+   * derived from the indices.
    *
-   * @param idx The indices for which to retrieve the preimage secret
-   * @return The preimage secret associated to the indices if it exists. Else None
+   * @param propositionEvidence The evidence of the Signature Proposition for which to retrieve the indices for
+   * @return The indices associated to the Signature Proposition evidence if it exists. Else a DataApiException
    */
-  def getPreimage(idx: Indices): Option[Preimage]
+  def getIndices(propositionEvidence: Evidence): F[Either[DataApiException, Indices]]
```
mundacho, line 57: What happens when the proposition evidence has multiple signatures??

mundacho (changes requested):
I put some comments that we need to discuss before merging.

brambl-sdk/src/main/scala/co/topl/brambl/dataApi/DataApi.scala
DiademShoukralla: The `propositionEvidence` is meant to only be for a Signature proposition, in which case, only a single signature and indices are associated with it. I wasn't sure how to enforce that the evidence should only be for the specific type of proposition, so I will update the `getIndices` and `getPreimage` functions to instead take in `Proposition.DigitalSignature` and `Proof.DigitalSignature` respectivefully.

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderInterpreter.scala
```diff
@@ -25,25 +25,19 @@ object TransactionBuilderInterpreter {
...
-          BuilderError
-            .InputBuilderError(
-              s"Could not construct input. Id=${data.address}, Attestation=${attestation}, Value=${value}"
-            )
-            .asLeft[SpentTransactionOutput]
-            .pure[F]
+    ): F[Either[BuilderError.InputBuilderError, SpentTransactionOutput]] =
+      (
+        for {
+          utxo <- EitherT(dataApi.getUtxoByTxoAddress(data.address))
```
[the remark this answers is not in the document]
DiademShoukralla: You are correct for LockAddresses that there may be multiple UTXOs, but for a TransactionOutputAddress I believe there should be only one UTXO since this type of address refers to a Transaction Output directly. 

However,  I think this function should be modified to take in either a TransactionOutputAddress or a LockAddress. I was hoping to do that change as a part of a separate ticket (perhaps repurposing TSDK-263)

mundacho (approved):

PR #54, opened 2023-04-28
TSDK-467 Encoding functions

mundacho:
## Purpose

The goal of this PR is to implement AddressCodec

## Approach

The implementation just transforms from lock address to a base58check encoded string and back again.

## Testing

Unit tests were added.

## Tickets

* TSDK-467


DiademShoukralla (approved):
LGTM!

PR #59, opened 2023-05-08
TSDK-474 Support for proving already partially proven transactions

DiademShoukralla:
## Purpose

Previously the credentialler overides the existing proofs in a transaction that it is attempting to prove. This prevents any multi-participant usage since after the first participant proves the transaction (into a partially proven transaction), the proofs from participant 1 would get erased when participant 2 goes to prove the partially proven transaction. This PR adds support for keeping any existing valid proofs in the transaction and will only prove a proposition if it was not previously proven.

## Approach

- When proving a proposition, supply the existing proof from the attestation. If it is invalid (not the same type or empty) then we will generate a new proof and return that. Else, return the existing proof
- Broke and modularized `getProof` into multiple smaller functions (one for each type of proposition) since the function was becoming large and unreadable.

## Testing

- Added a unit test to test a 2 person complex lock; at least one of each type of proposition including composites. This complex scenario tests that existing proofs do not get overriden, but the previously unprovable challenges can become proven at a later time.
- Made other changes to MockHelpers and other tests (ContainsSignable and TransactionCost) that were affected to ensure they still pass. Notably, Now the credentialler expects that the number of proofs in an attestation matches the number of propositions exactly. Previously it did not matter as much since the proofs would be overridden
- ran `checkPR` and ensured was successful

## Tickets
* Closes TSDK-474

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/wallet/CredentiallerInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/wallet/CredentiallerInterpreter.scala
@@ -51,115 +51,6 @@ object CredentiallerInterpreter {
       } yield if (vErrs.isEmpty) provenTx.asRight else vErrs.asLeft
 
     /**
-     * Return a Proof that will satisfy a Proposition and signable bytes, if possible.
-     * Any unprovable leaf (non-composite) Propositions will result in a [[Proof.Value.Empty]].
-     * Leaf/Atomic/Non-composite Propositions are: Locked, Digest, Signature, Height, and Tick
-     * If there are valid existing proofs for any leaf Propositions, they should not be overwritten.
-     *
-     * It may not be possible to retrieve a proof if
-     * - The proposition type is not yet supported
-     *      (not one of Locked, Digest, Signature, Height, Tick, Threshold, And, Or, and Not)
-     * - The secret data required for the proof is not available (idx for signature, preimage for digest)
-     * - The signature routine is not supported (not ExtendedEd25519)
-     *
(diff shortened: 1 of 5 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/wallet/CredentiallerInterpreter.scala
```diff
@@ -51,109 +51,273 @@ object CredentiallerInterpreter {
...
+     * @return The same input, but proven.
+     */
+    private def proveInput(
+      input: SpentTransactionOutput,
+      msg:   SignableBytes
+    ): F[SpentTransactionOutput] = {
+      val attestation: F[Attestation] = input.attestation.value match {
+        case Attestation.Value.Predicate(Attestation.Predicate(predLock, proofs, _)) =>
+          predLock.challenges
+            // TODO: Fix .getRevealed
```
mundacho, line 70: why is the TODO here?

mundacho (approved):
lgtm!

mundacho (approved):
lgtm!

brambl-sdk/src/main/scala/co/topl/brambl/wallet/CredentiallerInterpreter.scala
DiademShoukralla: Ah it looks like it was added in this PR since the whole function was moved up, but the TODO already existed previously (it was added in when the definition of Challenges expanded to include "PreviousProposition"). However we are not using/considering that type for now (only revealed type).

PR #63, opened 2023-05-17
TSDK-479 Implemented Lock and Proposition Templates

DiademShoukralla:
## Purpose

The y layer of the cartesian indexing scheme represents Contract/Lock Templates. These templates must be deterministic and portable to support the ability of sharing across participants. This PR adds the ability to define a new Lock (or proposition) template as well as the ability to instantiate a new instance by providing a list of VerificationKeys (which should map to the x layer of the cartesian indexing).

## Approach

- Added the ability to create Proposition Templates. Proposition Templates generally will not be used by the clients themselves; instead, they will create a Lock Template which contains Proposition templates within it. The following Proposition Templates are possible to be created:
    - Locked
    - Height
    - Tick
    - Digest
    - Signature
    - And
    - Or
    - Not
    - Threshold
- Added the ability to create a Lock Template. The only type of Lock we support are Predicate Locks where the Challenges are of type Revealed.
- Added the ability to instantiate a Lock or Proposition instance using it's respective template and a list of Verification Keys. As mentioned above, clients generally will not need to instantiate a Proposition from it's template, but instead instantiate a Lock (which contains proposition templates within).
-  Added JSON Serialization for Lock and Proposition Templates. The JSON representation should be portable and deterministically result in a Lock (or Proposition) template.
- Added tests for Building Templates/instantiating instances as well as serializing and deserializing the templates.

## Testing

- Added Tests for Building Lock Templates and instantiating Locks via a Template
- Added Tests for Building Proposition Templates and instantiating Propositions via a Template
- Added Tests for Serializing and Deserializing the JSON representation of Lock and Proposition Templates
- Ran `checkPR` and ensured all existing tests pass

## Tickets
* closes TSDK-479

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/locks/LockTemplate.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/locks/LockTemplate.scala
@@ -1,0 +1,39 @@
+package co.topl.brambl.builders.locks
+
+import co.topl.brambl.builders.BuilderError
+import co.topl.brambl.models.box.{Challenge, Lock}
+import LockTemplate.LockType
+import cats.Monad
+import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxEitherId, toFlatMapOps, toFunctorOps}
+import co.topl.brambl.builders.locks.PropositionTemplate.ThresholdTemplate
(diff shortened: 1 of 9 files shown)


mundacho (approved):
Looks good, can't wait to include it in Brambl-cli

PR #65, opened 2023-05-30
TSDK-489 Migrated Transaction Builder API + Other changes

DiademShoukralla:
## Purpose

The user of the BramblSc Transaction Builder is brambl-cli. The functions required by brambl-cli was initially created in brambl-cli. In this PR they were migrated to BramblSc + Additional changes (such as no longer hard coding the type of predicate lock; brambl-cli is using the templates)

## Approach

- Migrated API from brambl-cli
- Minor changes + bug fixes (bug fix in templates, removed some functions from the API, added condition in buildSimpleLvlTransaction to only add change output if value is >0)
- Added additional functions in WalletApi for key derivation and an additional function in WalletStateAlgebra to support other changes in brambl-cli
- Added unit tests 

## Testing

- Added additional unit tests
- Used in brambl-cli locally

## Tickets
* Part of TSDK-489


mundacho (approved):
👍

PR #70, opened 2023-06-19
TSDK-477 Separate DataApi into specific components (WalletApi and GenusQuery)

DiademShoukralla:
## Purpose

Instead of having a single DataApi, we will have specific data Apis within a dataApi package. This way, only the specific parts of the API that needs to be overridden can be. For example, users of BramblSc will need to define their own local storage mechanisms for Wallet State and Wallet Key APIs, but can all share the Genus Query implementation.

## Approach

- Moved WalletStateAlgebra into dataApi package
- Renamed DataApi into WalletKeyApiAlgebra (and remove functions no longer used)
- Migrated the GenusQueryAlgebra from brambl-cli into the dataApi package
- Updated any other parts of code affected by these changes

## Testing

- Ensured all existing tests are successful
- Used the changes in brambl-cli successfully

## Tickets
* TSDK-477


brambl-sdk/src/main/scala/co/topl/brambl/dataApi/GenusQueryAlgebra.scala
```diff
@@ -0,0 +1,49 @@
...
+   */
+  def queryUtxo(fromAddress: LockAddress): F[Seq[Txo]]
+}
+
+object GenusQueryAlgebra {
+
+  def make[F[_]: Sync](channelResource: Resource[F, ManagedChannel]): GenusQueryAlgebra[F] =
+    new GenusQueryAlgebra[F] {
+
+      def queryUtxo(fromAddress: LockAddress): F[Seq[Txo]] = {
```
nandotorterolo, line 30: state could be a parameter with default value unspent

nandotorterolo (approved):

brambl-sdk/src/main/scala/co/topl/brambl/dataApi/GenusQueryAlgebra.scala
DiademShoukralla: Thanks for the review! I made the change

PR #72, opened 2023-06-28
Update TransactionSyntaxInterpreter.scala

nandotorterolo:
## Purpose
fix scala doc

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
@@ -74,7 +74,7 @@ object TransactionSyntaxInterpreter {
     Validated.condNec(transaction.outputs.size < Short.MaxValue, (), TransactionSyntaxError.ExcessiveOutputsCount)
 
   /**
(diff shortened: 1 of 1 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -74,7 +74,7 @@ object TransactionSyntaxInterpreter {
     Validated.condNec(transaction.outputs.size < Short.MaxValue, (), TransactionSyntaxError.ExcessiveOutputsCount)
 
   /**
-   * Verify that the timestamp of the transaction is positive (greater than 0).  Transactions _can_ be created
+   * Verify that the timestamp of the transaction is positive (greater than or equal to 0).  Transactions _can_ be created
```
SeanCheatham, line 77: Thanks for the correction!  Would you mind also renaming the method to `nonNegativeTimestampValidation`? (Positive means "greater than zero", so this method is mis-named)

nandotorterolo: thank you for the review! I will rename it

DiademShoukralla (approved):

PR #50, opened 2023-04-26
[BN-997]: TDSK Deployment Tooling

admiraladmirable:
## Purpose
Enable publishing TSDK tooling to Maven.

## Approach
Enable the existing Maven release workflow, and add additional workflows following the [sbt-ci-release](https://github.com/sbt/sbt-ci-release/tree/main#github-actions) repo.

High level overview:
* Each PR merged to main triggers the [release-drafter](https://github.com/marketplace/actions/release-drafter) workflow which creates a draft release and/or adds a link to the PR under a changes section.
* When we are ready to release, publish the draft which will push tags to the main repo.
* The Maven Release workflow triggers when tags are pushed to main, which publishes

## Testing
* Ran the Maven publish workflow targeting my branch: https://github.com/Topl/BramblSc/actions/runs/4813348677

I can't test the release drafter workflow since it requites the `./github/release-drafter.yml` file to exist in the main branch. I will verify this works after the changes are merged.

## Tickets
* closes [BN-997](https://topl.atlassian.net/browse/BN-997) *

[BN-997]: https://topl.atlassian.net/browse/BN-997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #73, opened 2023-06-30
TSDK-496 Prepare to Deploy to Maven

DiademShoukralla:
## Purpose

In preparation to deploy BramblSc to maven, some changes were made.

## Approach

 The tooling was already added in https://github.com/Topl/BramblSc/pull/50 , however this PR contains minor changes; prefixing `sonatypeCredentialHost` with `ThisBuild / ` and the `https://s01.oss.sonatype.org/content/repositories/releases/` resolver.

Additionally, the Topl dependencies (protobuf-specs and quivr4s) were also updated from jitpack to the released Maven artifacts. 

## Testing

- ensured released Topl dependencies from Maven are being used
image
- ran unit tests in both projects and ensured successful (sanity)
- ran `sbt checkPR` and ensured successful (sanity)
- ran `sbt publishLocal` and ensured successful (sanity)

The deployment will be tested after this PR is merged in and the 2.0.0-alpha1 tag is pushed

## Tickets
* TSDK-496

### DIFF
--- a/.github/workflows/maven_release.yml
+++ b/.github/workflows/maven_release.yml
@@ -24,4 +24,3 @@ jobs:
           PGP_SECRET: ${{ secrets.PGP_SECRET }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
-          CI_SONATYPE_RELEASE: ${{ secrets.CI_SONATYPE_RELEASE }}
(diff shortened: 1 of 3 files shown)


admiraladmirable (approved):
I'm curious if we still need to explicitly add `CI_SONATYPE_RELEASE`, or if we can get away with the defaults. For now, I think this is fine, but I think we should make the Maven releases relatively consistent for our different repos.

DiademShoukralla:
@admiraladmirable Thanks for the review! I absolutely agree; I will remove CI_SONATYPE_RELEASE and see if we can get away with the default

PR #75, opened 2023-07-06
TSDK-530 Minor fixes to Locked and Digest Proposition Templates

DiademShoukralla:
## Purpose

There was a bug in Locked Proposition templates where instead of storing the value of the Locked data, we stored the string representation of the bytestring.
 In addition, the digest within the Digest Proposition Template was not human readable.

## Approach

- For encoding Locked to JSON, instead of using `a.data.get.value.toString`, we use `encodeToBase58(a.data.get.value.toByteArray)`
- For decoding Locked from JSON, we `decodeFromBase58` and then convert the decoded data into Locked Proposition. If decoding the data fails, we use a Custom Decoding (circe's DecodingFailure) reason
- For encoding Digest to JSON, instead of using `Strings.fromByteArray`, we use `encodeToBase58`
- For decoding Digest from JSON, we parse the digest value then `decodeFromBase58`. If successful we convert the decoded data into Digest Proposition. If it failed, we use a Custom Decoding (circe's DecodingFailure) reason

## Testing

- Updated MockData that is used in the unit tests. Previously we didnt test the optional data within the Locked proposition (which is why it was not caught). Ensured all tests pass
image
- Ran `checkPR`
- Printed the new JSONs manually to ensure it appears in base58: (locked and digest)
image
image

## Tickets
* Closes TSDK-530

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/codecs/PropositionTemplateCodecs.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/codecs/PropositionTemplateCodecs.scala
@@ -1,10 +1,12 @@
 package co.topl.brambl.codecs
 
 import cats.Monad
-import cats.implicits.catsSyntaxEitherId
+import cats.implicits.{catsSyntaxEitherId, catsSyntaxOptionId, toBifunctorOps}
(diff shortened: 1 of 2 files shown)


mundacho (approved):
LGTM!

PR #78, opened 2023-07-12
Updated README.md with updated consumption instructions

DiademShoukralla:
## Purpose

- In addition to Jitpack, BramblSc can be consumed via maven releases. The README should be updated to reflect that
- This will serve as a test to https://github.com/Topl/BramblSc/pull/77

## Approach

- updated README.md

## Testing

n/a

## Tickets
* No ticket
* Tests TSDK-507

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,11 @@
 
 Topl's Brambl SDK implemented in Scala
 
-Multiple artifacts will be built from this repo. Some will be just for Topl clients and some will be shared. Currently 
+Multiple artifacts will be built from this repo. Some will be just for Topl clients and some will be shared. 
 
-This repo should be consumed using jitpack. Here is how:
+## Consume with JitPack
+
+This repo can be consumed using jitpack. Here is how:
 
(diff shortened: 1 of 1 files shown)


nandotorterolo (approved):
Looks good!  if you can remove the sort operation in the budget would be great:
line 1
> https://img.shields.io/github/v/tag/topl/bramblsc?label=release

- My understanding is that 2.0.3 > 2.0.0-alpha1. and for this reason the budget is wrong
- Not sure why if the is a 2.0.3 version already there, why we are now in 2.0.0-alpha1

DiademShoukralla:
@nandotorterolo Thanks for the review! I made the change. Yeah, it looks like 0.0.1, 0.0.2, 2.0.2, 2.0.3 we're created in January as part of deployment tests, and then was disabled. It looks like these tags are not used/were not published to Maven: https://s01.oss.sonatype.org/content/repositories/releases/co/topl/ so I will delete them on github

PR #86, opened 2023-08-01
BN-1138/113434 group and series asset token

nandotorterolo:
## Purpose

TAMv2 Assets
A TAMv2 asset is a token that has the following attributes:

- A group identifier
- A series identifier
- A quantity

Asset Lifecycle

- Mint a group constructor token
- Mint a series constructor token (this can be done in the same transaction as the group constructor is minted)
- Mint an asset token using the group and series constructor token just minted.

### Note

This PR is linked to the following PRs, and only group constructor token, with empty series was implemented.

- https://github.com/Topl/protobuf-specs/pull/78
- https://github.com/Topl/BramblSc/pull/86

## Approach
- implement Validation defined on https://github.com/Topl/tips/tree/main/TIP-0003

## Testing
- preparePR

## Tickets
https://topl.atlassian.net/browse/BN-909


brambl-sdk/src/test/scala/co/topl/brambl/validation/TransactionSyntaxInterpreterSpec.scala
```diff
@@ -174,4 +179,102 @@ class TransactionSyntaxInterpreterSpec extends munit.FunSuite with MockHelpers {
       .exists(_.toList.contains(TransactionSyntaxError.InvalidDataLength))
     assertEquals(result, true)
   }
+
+  test("validate distinct output groups") {
```
nandotorterolo, line 183: it should be possible

nandotorterolo: removed

brambl-sdk/src/main/scala/co/topl/brambl/syntax/GroupPolicySyntax.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  implicit def groupPolicyAsGroupPolicySyntaxOps(groupPolicy: GroupPolicy): GroupPolicyAsGroupPolicySyntaxOps =
+    new GroupPolicyAsGroupPolicySyntaxOps(groupPolicy)
+}
+
+class GroupPolicyAsGroupPolicySyntaxOps(val groupPolicy: GroupPolicy) extends AnyVal {
+
+  def computeId: GroupId = {
+    val digest: Array[Byte] = groupPolicy.immutable.value.toByteArray
+    val sha256 = MessageDigest.getInstance("SHA-256").digest(digest)
```
mundacho, line 21: I am not sure, but I think sometime Blake2b is used as hash, we might need to consult with Diadem or with Hans

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -247,4 +263,79 @@ object TransactionSyntaxInterpreter {
...
+
+    def utxoIsPresent(addresses: Seq[TransactionOutputAddress]) =
+      transaction.inputs.exists(spentTxOutput =>
+        addresses.contains(spentTxOutput.address) &&
+        spentTxOutput.value.value.isLvl
+      )
+
+    val validations = (transaction.groupPolicy, transaction.seriesPolicy) match {
+      // there are group/series constructors tokens, and policies
+      case (groupPolicies, seriesPolicies) if groupPolicies.nonEmpty && seriesPolicies.nonEmpty =>
```
mundacho, line 303: I don't think that the cases are needed. You can just do all the validations for both group and policies and it will still work.

mundacho (approved):
Just one comment that we can discuss later, but otherwise looks good to me.

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
nandotorterolo: thanks

brambl-sdk/src/main/scala/co/topl/brambl/syntax/SeriesPolicySyntax.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  implicit def seriesPolicyAsSeriesPolicySyntaxOps(SeriesPolicy: SeriesPolicy): SeriesPolicyAsSeriesPolicySyntaxOps =
+    new SeriesPolicyAsSeriesPolicySyntaxOps(SeriesPolicy)
+}
+
+class SeriesPolicyAsSeriesPolicySyntaxOps(val seriesPolicy: SeriesPolicy) extends AnyVal {
+
+  def computeId: SeriesId = {
+    val digest: Array[Byte] = seriesPolicy.immutable.value.toByteArray
+    val sha256 = MessageDigest.getInstance("SHA-256").digest(digest)
```
nandotorterolo, line 21: I am not sure, but I think sometime Blake2b is used as hash, we might need to consult with Diadem or with Hans

brambl-sdk/src/main/scala/co/topl/brambl/syntax/GroupPolicySyntax.scala
nandotorterolo: - SHA-256 is widely used in the Bitcoin blockchain, e.g. for identifying the transaction hashes and for the proof-of-work mining performed by the miners.
- SHA3-256 provides more cryptographic strength than SHA-256 for the same hash length (256 bits).
- The hash function Keccak-256, which is used in the Ethereum blockchain, is a variant of SHA3-256 with some constants changed in the code.
- BLAKE2s (typically 256-bit) is BLAKE2 implementation, performance-optimized for 32-bit microprocessors.
-As recommendation, prefer using SHA-2 and SHA-3 instead of RIPEMD, because they are more stronger than RIPEMD, due to higher bit length and less chance for collisions.
-No Collisions for SHA-256, SHA3-256, BLAKE2s and RIPEMD-160 are Known 
- https://wizardforcel.gitbooks.io/practical-cryptography-for-developers-book/content/cryptographic-hash-functions/secure-hash-algorithms.html

DiademShoukralla: From the perspective of what we are currently using, we are using Blake2b to compute evidences, to bind the transaction binding to the proofs, and is currently our hashing routine used for Digest Propositions/Proofs (as defined in Brambl's Dynamic Context). 
SHA is used throughout the crypto library; SHA256 is used for mnemonics and SHA512 is used for key generation and ExtendedEd25519. SHA256 is also used for the checksum in our base58's encoding/decoding

DiademShoukralla (approved):

brambl-sdk/src/main/scala/co/topl/brambl/syntax/GroupPolicySyntax.scala
nandotorterolo: Thanks for your review. Just for context, in the last sprint review, we define with Hans that we should proceed with Sha-256 and then implement this ticket. Wich talks about "Topl should standardize the hashing method in the whole ecosystem."
https://topl.atlassian.net/browse/BN-1160

PR #89, opened 2023-08-03
TSDK-537 Import Quvir4s

DiademShoukralla:
## Purpose

To simplify our dev and release processes, we want to integrate Quivr4s into the BramblSc repo. 

## Approach

- Added the quivr4s implementation from https://github.com/Topl/quivr4s as a sub project in this repo
- A notable change from the implementation in the quivr4s repo is the use of blake2b256 hash. Previously the function was duplicated: https://github.com/Topl/quivr4s/blob/main/src/main/scala/co/topl/quivr/api/package.scala . I replaced the usages of this with the blake2b hash from crypto (same implementation). The reason it was not used before was to prevent cyclic dependency across repos (using JitPack), but that is no longer an issue.
- Removed the old quivr4s import

## Testing

- ran checkPR and ensured was successful
- Quivr4s tests were added as well. I ensured that they run as part of checkPR

## Tickets
* TSDK-537


DiademShoukralla:
@nandotorterolo Thanks for the review! I think you're right that it was for the initial POC; I'm in favor of deleting them too. I will make the change

mundacho (approved):
LGTM!

PR #98, opened 2023-08-31
TSDK-572 Address WalletApi flaky test with CatsUnsafeResource

DiademShoukralla:
## Purpose

there is a [flaky test in the wallet API](https://github.com/Topl/BramblSc/actions/runs/5809330070/job/15747898079#step:8:271) that fails with parallel execution
image. It is caused by using an implicit ExtendedEd25519 in the WalletApi. 

## Approach

Instead of using implicit ExtendedEd25519, use a CatsUnsafeResource queue to manage its use.

Note: Adding CatsUnsafeResource (and updating WalletApi_ was a small change, most of the changes in this PR is updating the tests to align with the changes to WalletApi (Replacing Id[_] with IO[_]).

## Testing

- Ran tests locally 20x in a row with parallel execution => flaky test did not fail
- Ran tests locally 20x in a row without parallel execution => flaky test did not fail
- Temporarily updated the dependencies in bifrost and ensured all checks pas
- Locally updated the dependencies in brambl-cli and ensured all the unit tests pass

## Tickets
* TSDK-550
* TSDK-572

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/wallet/WalletApi.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/wallet/WalletApi.scala
@@ -1,8 +1,8 @@
 package co.topl.brambl.wallet
 
-import cats.implicits.catsSyntaxApplicativeId
+import cats.implicits.{catsSyntaxApplicativeId}
 import co.topl.crypto.generation.mnemonic.{Entropy, MnemonicSize, MnemonicSizes}
-import cats.Monad
+import cats.{Eq, Monad}
 import cats.implicits.{toFlatMapOps, toFunctorOps}
 import co.topl.brambl.dataApi.WalletKeyApiAlgebra
 import co.topl.crypto.generation.{Bip32Indexes, KeyInitializer}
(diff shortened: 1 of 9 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/wallet/WalletApi.scala
```diff
@@ -285,14 +291,15 @@ object WalletApi {
    * @param walletKeyApi The Api to use to handle wallet key persistence
    * @return A new WalletAPI instance
    */
-  def make[F[_]: Monad](
-    walletKeyApi: WalletKeyApiAlgebra[F]
-  )(implicit extendedEd25519Instance: ExtendedEd25519 = new ExtendedEd25519): WalletApi[F] = new WalletApi[F] {
+  def make[F[_]: Monad: Async](walletKeyApi: WalletKeyApiAlgebra[F]): WalletApi[F] = new WalletApi[F] {
```
mundacho, line 294: Async is already a Monad, in the code you just needed to replace Monad[F] by Async[F]

mundacho (approved):
Although this code is very interesting and it makes any resource thread safe. I think for the case in question (the ExtendedEd25519), just doing a new each time should solve the problem, as the problem was reusing the same global resource each time.

PR #99, opened 2023-09-01
TSDK-568 Group Constructor Minting Transaction

DiademShoukralla:
## Purpose

Add capability to TransactionBuilder to create a Group Minting transaction.

## Approach

- Added function `buildSimpleGroupMintingTransaction` to TransactionBuilder. This function takes in WalletStateAlgebra and BifrostQueryAlgebra to query data from the wallet state and a bifrost node respectively, a group Policy (for which to mint constructor tokens for), a lock (to encumber the newly minted tokens), and quantity of tokens to mint (+optional change information). WalletStateAlgebra and BifrostQueryAlgebra are needed to retrieve information on the registrationUtxo specified in the groupPolicy.
- Added WalletStateAlgebra.getLockByAddress to retrieve the lock of the registrationUtxo from wallet state. The bifrost call (fetchTransaction) was already implemented

## Testing

- Added tests for buildSimpleGroupMintingTransaction and getLockByAddress. The former required changes to TransactionBuilderInterpreterSpec to switch Id[_] to IO[_]
- ensured all existing tests pass

## Tickets
* Closes TSDK-568


brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
```diff
@@ -150,6 +184,70 @@ object TransactionBuilderApi {
           .withDatum(datum)
       } yield ioTransaction
```
mundacho, line 186: These are very heavy objects to pass around (walletStateApi and bifrostRPC), I would prefer we pass directly whatever is needed instead of computing this here. Also, calling methods from walletState and bifrostRPC adds the possibility of having new different errors. For example, network and database issues. My suggestion is to move (in a new task if necessary) all the code that needs theses two objects to the service-kit

mundacho (changes requested):
It looks good to me except for the comment I left. We can make a call and discuss it if you think there is no way around. I saw that some tests require testing for example that the UTXO exist. Maybe another solution is to move this transaction builder entirely to the service kit. What do you think?

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
DiademShoukralla: As discussed, I will remove the (walletStateApi and bifrostRPC) parameters and replace with whatever is needed directly (TXO, Lock.Predicate). I will also add user validation to ensure proper use of the function

brambl-sdk/src/test/scala/co/topl/brambl/builders/TransactionBuilderInterpreterSpec.scala
```diff
@@ -19,7 +23,7 @@ class TransactionBuilderInterpreterSpec extends munit.FunSuite with MockHelpers
       trivialLockAddress,
       1
     )
-    assert(testTx.computeId == txFull.computeId)
+    assertIO(testTx.map(_.computeId), txFull.computeId)
```
mundacho, line 26: I think that you do not need the IO anymore if we are not passing the walletState and bifrostQuery

mundacho (approved):
LGTM

brambl-sdk/src/test/scala/co/topl/brambl/builders/TransactionBuilderInterpreterSpec.scala
DiademShoukralla: You are correct, I have removed those changes

PR #103, opened 2023-09-11
TAM v2 models

nandotorterolo:
## Purpose
- update protobuf-specs with latest TAM v2 models
- remove validations on Asset, they will be implemented in other PR


DiademShoukralla (approved):
Thanks for this!

PR #104, opened 2023-09-12
TSDK-570 Add Asset minting to Transaction builder

DiademShoukralla:
## Purpose

Added the ability to create a transaction to mint TAMv2 Asset tokens

## Approach

- Added new function signature `buildSimpleAssetMintingTransaction`
- Refactored the User Input validations to reduce code reuse (certain validations are common). Note that validating the metadata is a nice-to-have and will be implemented in a future ticket (TSDK-585)
- Added tests

## Testing

- updated the existing tests for minting Group and Series constructor tokens to align with the changes to User input validation
- Identified and added 18 new test cases for asset minting. These cover both success and failure cases.

## Tickets
* Closes TSDK-570

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
@@ -344,7 +344,7 @@ object TransactionBuilderApi {
           stxoAttestation <- EitherT.right[BuilderError](unprovenAttestation(registrationLock))
           datum           <- EitherT.right[BuilderError](datum())
           utxoMinted <- EitherT.right[BuilderError](
-            groupOutput(mintedConstructorLockAddress, quantityToMint, groupPolicy.computeId)
+            groupOutput(mintedConstructorLockAddress, quantityToMint, groupPolicy.computeId, groupPolicy.fixedSeries)
           )
         } yield IoTransaction(
           inputs = Seq(
@@ -449,7 +449,12 @@ object TransactionBuilderApi {
             )
           )
           groupOutput <- EitherT.right[BuilderError](
-            groupOutput(groupTxo.transactionOutput.address, groupToken.quantity, groupToken.groupId)
+            groupOutput(
+              groupTxo.transactionOutput.address,
+              groupToken.quantity,
+              groupToken.groupId,
+              groupToken.fixedSeries
+            )
           )
           seriesOutputSeq <- {
             val inputQuantity = BigInt(seriesToken.quantity.value.toByteArray)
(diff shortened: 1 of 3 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
```diff
@@ -146,7 +193,81 @@ object TransactionBuilderApi {
...
+      )
+
+    def validMintingSupply(
+      testQuantity:   Int128,
+      seriesTokenOpt: Option[Series],
+      testLabel:      String
+    ): ValidatedNec[UserInputError, Unit] =
+      (testQuantity, seriesTokenOpt) match {
+        case (d, Some(Series(_, a, Some(tokenSupply), _, _, _))) =>
+          val desiredQ = BigInt(d.value.toByteArray)
```
mundacho, line 261: Not sure if you do somewhere else, but we need to check d > 0. Otherwise, desiredQ % tokenSupply == 0 can also happen, but still it is invalid.

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
```diff
@@ -146,7 +193,81 @@ object TransactionBuilderApi {
...
+          val availableQ = BigInt(a.value.toByteArray)
```
mundacho, line 262: Same thing applies here. availableQ == 0 will make desiredQ > availableQ * tokenSupply always. Be sure to check here or somewhere else.

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
```diff
@@ -282,53 +410,243 @@ object TransactionBuilderApi {
...
+        groupId:                GroupId,
+        seriesId:               SeriesId,
+        fungibilityType:        FungibilityType,
+        quantityDescriptorType: QuantityDescriptorType,
+        metadata:               Option[Struct],
+        commitment:             Option[ByteString]
+      ): F[UnspentTransactionOutput] = {
+        val groupSeriesValues = fungibilityType match {
+          case FungibilityType.GROUP_AND_SERIES => (groupId.some, seriesId.some, None, None)
+          case FungibilityType.SERIES =>
```
mundacho, line 618: We do not plan to support those yet. I think maybe we could set an unsupported exception here.

mundacho (approved):
Looks good to me. The only thing that may cause an issue is the fungibility. For the first version (not TGE, but first internal version) we would like to only support group and series fungibility.  Since it is internal, we can leave it there, how much work is it to remove that part?

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
DiademShoukralla: This check happens somewhere else (line 560, "positiveQuantity")

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
DiademShoukralla: Good catch! I will add that

brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
DiademShoukralla: Ah thanks for letting me know! I will make the change

PR #106, opened 2023-09-13
Validations_Tam_v2_Well-formedness_A

nandotorterolo:
## Purpose

 Well-formedness validation
- Only for each asset: a input assets + a minted assets == total output of a assets

## Approach
- A new validation rule was created
- A new specific test was created, only to evaluate this rule.

## Tickets
- ???

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
@@ -8,7 +8,6 @@ import co.topl.brambl.models.transaction.IoTransaction
 import co.topl.brambl.validation.algebras.TransactionSyntaxVerifier
 import co.topl.brambl.common.ContainsImmutable.ContainsImmutableTOps
 import co.topl.brambl.common.ContainsImmutable.instances._
-import co.topl.brambl.models.TransactionOutputAddress
 import co.topl.brambl.models.box.Attestation
 import co.topl.brambl.syntax._
 import quivr.models.{Int128, Proof, Proposition}
@@ -37,7 +36,8 @@ object TransactionSyntaxInterpreter {
       positiveOutputValuesValidation,
       sufficientFundsValidation,
       attestationValidation,
-      dataLengthValidation
+      dataLengthValidation,
+      assetEqualFundsValidation
     )
 
   /**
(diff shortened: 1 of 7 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -232,4 +233,72 @@ object TransactionSyntaxInterpreter {
...
+        .filter(sto => transaction.mintingStatements.map(_.groupTokenUtxo).contains(sto.address))
+        .map(_.value.getGroup.groupId)
+
+    val seriesIdsOnMintedStatements =
+      transaction.inputs
+        .filter(_.value.value.isSeries)
+        .filter(sto => transaction.mintingStatements.map(_.seriesTokenUtxo).contains(sto.address))
+        .map(_.value.getSeries.seriesId)
+
+    val mintedAsset = transaction.outputs
```
mundacho, line 258: This is not right, the number of total assets minted should come from the asset minting statement (that has the quantity in it.)

mundacho (changes requested):

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
nandotorterolo: Thanks for reviewing it, and explanation.

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -232,4 +232,76 @@ object TransactionSyntaxInterpreter {
...
+        .headOption
+
+    val mintedAsset = transaction.mintingStatements.map { stm =>
+      Value.defaultInstance
+        .withAsset(
+          Value.Asset(
+            groupId = groupGivenMintedStatements(stm).map(_.groupId),
+            seriesId = seriesGivenMintedStatements(stm).map(_.seriesId),
+            quantity = stm.quantity
+          )
```
DiademShoukralla, line 266: I think the `fungibility` field also needs to be set here. The reason is that the grouping is also based on the Asset fungibility (`value.typeIdentifier` on line 274 will return something different and thus be compared to the wrong assets in the outputs). 

You can get the fungibility type from `seriesGivenMintedStatements(stm).map(_.fungibility).getOrElse(GROUP_AND_SERIES)` (i.e, defaulting to GROUP_AND_SERIES)

DiademShoukralla (approved):

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
nandotorterolo: thanks for the suggestion, and provide the code implementation,  I will include it and a new test coverage

mundacho (approved):
Minor correction as discussed

PR #107, opened 2023-09-14
TSDK-578 Transfer Tokens

DiademShoukralla:
## Purpose

Transfer tokens while considering different token types. Specifically, The user can transfer X amount of a specific type of token to a recipient address while providing ALL tokens at the same LockAddress. Any tokens not going to the recipient will go to a change address

## Approach

- Added `buildLvlTransferTransaction`, `buildGroupTransferTransaction`, `buildSeriesTransferTransaction`, and `buildAssetTransferTransaction` to the Transaction Builder API
- All 4 of these functions under-the-hood will use private function `buildTransferTransaction` which handles the actual implementation (and takes in addition a `ValueTypeIdentifier` to specify which token is meant to be transferred to the recipient)
- At a high level, this function 
    - Expects all provided Txos to correspond with the provided `lockPredicateFrom`. With this lock predicate, we create the attestation for all inputs. Each provided Txo becomes an input
    - To create the UTXOs, we first merge all Txos of the same type (i.e, same group ID if type is Group Constructor Token). Then we retrieve the merged TXO that corresponds to the type we want to transfer. With this, if the quantity exceed the amount to transfer, we split it into 2. With all of these values, we turn them into UTXOs with their corresponding LockAddresses.
- Made the existing functions `groupOutput`, `seriesOutput`, and `assetOutput` public (to align with the `lvlOutput` function)
- Minor changes (that do not affect their existing algorithms) to some existing functions in TransactionBuilder to use new SyntaxOps which reduces the verbosity of their code
- Moved `UserInputValidations` into its own file + added validations specifically for transfer
-  Added tests

## Testing
- Added tests for the 4 new functions `buildLvlTransferTransaction`, `buildGroupTransferTransaction`, `buildSeriesTransferTransaction`, and `buildAssetTransferTransaction`. We test user input failures and 1 complex (merge and split) case and 1 simple (single input) case
- Ran tests in brambl-cli on the main branch to ensure backwards compatibility.
    - Added the stubs for the newly added functions
    - ran unit tests and integrations tests. ensured all successful

## Tickets
* closes TSDK-578

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/BuilderError.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/BuilderError.scala
@@ -7,3 +7,7 @@ package co.topl.brambl.builders
  * @param message The error message
  */
 abstract class BuilderError(message: String, cause: Throwable = null) extends RuntimeException(message, cause)
+
+case class UserInputError(message: String) extends BuilderError(message, null)
+
+case class BuilderRuntimeError(message: String, cause: Throwable = null) extends BuilderError(message, cause)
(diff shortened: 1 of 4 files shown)


mundacho (commented):
It looks good to me, but I think we need the fees. what do you think about that?

DiademShoukralla:
@mundacho Thanks for the review! I made the changes

mundacho (approved):
LGTM!

PR #108, opened 2023-09-15
BN-1200 well-formedness-b-no-repeated-utxos

nandotorterolo:
## Purpose

Description

Well-formedness

- **B**: For all assets minting statement ams1, ams2, etc => size(Set(ams1.utxo Union ams2.utxo .. amsn.utxo)) == size(List(ams1.utxo ..)  (No repeated UTXOs)

## Approach

Implement **rule B**, based on the branch, which contains Rule A: 

A:  Only for each asset: a input assets + a minted assets == total ouput of a assets

Once approved, this branch, should be merged with stepts
- A  <- B
- main <- A

https://topl.atlassian.net/wiki/spaces/LABS/pages/678658061/Validation+TAMv2+Transactions+-+Meeting+Fernando+Edmundo+2023-09-08

## Testing
unit test and preparePR

## Tickets
https://topl.atlassian.net/browse/BN-1201


mundacho (approved):
LGTM!

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -301,4 +302,25 @@ object TransactionSyntaxInterpreter {
...
+  private def assetNoRepeatedUtxosValidation(transaction: IoTransaction): ValidatedNec[TransactionSyntaxError, Unit] =
+    NonEmptyChain
+      .fromSeq(
+        transaction.mintingStatements
+          .flatMap(stm => Seq(stm.groupTokenUtxo, stm.seriesTokenUtxo))
+          .groupBy(identity)
+          .collect {
+            case (address, seqAddresses) if seqAddresses.size > 1 =>
+              TransactionSyntaxError.DuplicateInput(address)
+          }
```
DiademShoukralla, line 321: As discussed in call, this should take into account minting statements whose (groupUtxo, seriesUtxo) tuple is unique, but one of two addresses may be repeated. For instance, consider

```
MS1:
	groupUtxo: A
	seriesUtxo: B

MS2:
	groupUtxo: C
	seriesUtxo: B

((A, B), (C, B)) => (A, B, C, B) => (A: (A), C: (C), B: (B, B))
which would fail
```

cc: @mundacho

DiademShoukralla (approved):
Other than the comment I left, LGTM!

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
DiademShoukralla: Discussed with Edmundo, this will not be a case we will support. If user's want to mint MS2, they will have to do it in a subsequent transaction (where seriesUtxo points to the change in this Transaction)

mundacho: As discussed, that case should not be possible according to the spec. More specifically, as a design principle, we are going here with more strict but simple to validate constraints, than preferring more flexible but difficult to validate.

nandotorterolo:
ready to merge in branch [Validations_Tam_v2](https://github.com/Topl/BramblSc/tree/Validations_Tam_v2)

PR #113, opened 2023-09-21
TSDK-590 Support GROUP and SERIES fungibility + consider alloys and quantity descriptors for identifying a token type

DiademShoukralla:
## Purpose

To support GROUP and SERIES fungibility and to update the Token Identifier to take into consideration alloy assets and quantity descriptors.

## Approach

High level: 
- Updated TransactionBuilder + tests to support all types of fungibility. 
- A unique "kind" of asset token is one with also same [Group/Series] alloy (or ID if asset is not an alloy) as well as same Quantity Descriptor

Notable Callouts:
- Non-Alloy asset tokens require both GroupId and SeriesId. Only when it is an alloy asset token can we ignore one of those 2 fields. If not, we cannot construct the resulting alloy when merging 
- We are currently only supporting LIQUID quantity descriptor types in this PR. Temporary user validation was added to ensure any other types (IMMUTABLE, FRACTIONABLE, ACCUMULATOR) of asstets are not provided. The reason is these types have special behavior w.r.t aggregating and deagregating (impacts how we derive the change outputs). These quantity descriptor types will be addressed in future tickets:  TSDK-597, TSDK-598, and TSDK-599

## Testing

- Ensure existing tests pass
- Updated Transfer and Asset minting tests to incorporate GROUP only and SERIES only assets (success case)
- Added tests to ensure we are only supporting LIQUID quantity descriptor type for now.
- Tested grouping by "token types"

## Tickets
* Closes TSDK-590


brambl-sdk/src/main/scala/co/topl/brambl/builders/UserInputValidations.scala
```diff
@@ -123,30 +121,24 @@ object UserInputValidations {
...
-  def allValidFungibilityTypes(testValues: Seq[Value]): ValidatedNec[UserInputError, Unit] =
+  def allValidQuantityDescriptorTypes(testValues: Seq[Value]): ValidatedNec[UserInputError, Unit] =
     Validated.condNec(
-      // Any asset tokens must have valid fungibility
-      testValues.forall(v => !v.isAsset || v.asset.get.fungibility == FungibilityType.GROUP_AND_SERIES),
+      // Any asset tokens must have valid QuantityDescriptorType
+      testValues.map(_.typeIdentifier).map(identifierQuantityDescriptorType).forall(_.isValid),
       (),
-      UserInputError(s"All asset tokens must have valid fungibility type. We currently only support GROUP_AND_SERIES")
+      UserInputError(s"All asset tokens must have valid QuantityDescriptorType. We currently only support Liquid")
```
nandotorterolo, line 141:
```suggestion
      UserInputError(s"All asset tokens must have valid QuantityDescriptorType. We currently only support LIQUID")
```

brambl-sdk/src/main/scala/co/topl/brambl/syntax/TokenTypeIdentifierSyntax.scala
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * A Series fungible asset type, identified by a SeriesId, a Group alloy, and a QuantityDescriptorType. If the asset is
+ * not an alloy, the group "alloy" is given by the groupId.
+ *
+ * @param series The SeriesId of the asset
+ * @param groupAlloyOrId If the asset is an alloy, the Group alloy. Else the GroupId of the asset
+ * @param qdType The QuantityDescriptorType of the asset
+ */
+case class SeriesFungible(series: SeriesId, groupAlloyOrId: ByteString, qdType: QuantityDescriptorType)
```
nandotorterolo, line 109: In GroupFungible,  id was named "groupId", maybe here you can do the same?
```suggestion
case class SeriesFungible(seriesId: SeriesId, groupAlloyOrId: ByteString, qdType: QuantityDescriptorType)
```

brambl-sdk/src/main/scala/co/topl/brambl/builders/UserInputValidations.scala
nandotorterolo: I expect that this change will not create lot of failure tests,

nandotorterolo (approved):

brambl-sdk/src/main/scala/co/topl/brambl/syntax/TokenTypeIdentifierSyntax.scala
DiademShoukralla: Good catch! I will make the changes

brambl-sdk/src/main/scala/co/topl/brambl/builders/UserInputValidations.scala
DiademShoukralla: I will make the change and update the tests to match

mundacho (approved):
LGTM!

PR #114, opened 2023-09-21
TAMv2_Minting_Asset_Case3_step1

nandotorterolo:
## Purpose
- Validations only for minting (After projection to min (only if for all inputs and outputs isMint == true))
- Case 3: Given some asset a that is part of the transaction
- asset minting statement (ams) attached for asset a attached to transaction

## Approach

I considered  it should **NOT** be implemented, After doing the projection output assets are empty and I don't have elements to validate


nandotorterolo:
@mundacho closing

PR #115, opened 2023-09-21
Quick Fix: Preserve Transaction fields in Credentialler

DiademShoukralla:
## Purpose

Signature validation was failing due to the transaction signature changing after using the Credentialler to prove a transaction. The cause was that we were stripping all other fields (other than inputs, outputs, and datum) on the transaction. This was not caught before since we previously did not have any other fields on a transaction. The issue arose since we are now introducing Policies and Statements into the Tx.

## Approach

Instead of recreating the Proven Transaction using IoTransaction.defaultinstance (and populating some fields), we copy the unproven transaction and only override the inputs.

## Testing

- Ensured existing tests pass
- Added a new test case to check preserving fields

## Tickets
* N/a


mundacho (approved):
LGTM

PR #116, opened 2023-09-22
BN-1204-validations-only-for-minting-case-3_step_3

nandotorterolo:
## Purpose

Implement **case3**, 

Validations only for minting (After projection to min (only if for all inputs and outputs isMint == true))
- asset minted correspond to token supply in series policy
- It is not being considered series outputs value, because they will be implemented on Transfer-Series-Funds validation

https://topl.atlassian.net/wiki/spaces/LABS/pages/678658061/Validation+TAMv2+Transactions+-+Meeting+Fernando+Edmundo+2023-09-08

## Testing
unit test and preparePR

- add test for valida and invalid use case
- add visual diagram  valid number 5
image

- include invalid case number:6
image

## Tickets
https://topl.atlassian.net/browse/BN-1201


brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,7 +393,26 @@ object TransactionSyntaxInterpreter {
...
+        )
+        .map(_.value.getSeries)
+        .headOption
+
+    val validAssets = assets.forall { a =>
+      a.seriesId.isDefined &&
+      seriesIdOnMintedStatements(a).exists(s =>
+        s.tokenSupply match {
+          case Some(tokenSupplied) =>
+            (s.quantity: BigInt) * (tokenSupplied: BigInt) % (a.quantity: BigInt) == 0
```
DiademShoukralla, line 411: I'm not sure if this is correct. 

Consider the following case where `s.quantity = 2` and `tokenSupply = 3` and `a.quantity = 2`. The check would pass but I believe it should be invalid since `a.quantity` is not a multiple of `tokenSupply`. Thus failing the rule "asset minted correspond to token supply in series policy"

Perhaps we need to check if the quantity is a multiple of token supply separately from checking there is enough total supply. Maybe something like
```
tokenSupplied % a.quantity == 0 // Is minted quantity a multiple of token supply?
&&
s.quantity * tokenSupplied >= a.quantity  // Is there enough series supply to mint asset quantity?
```

What do you think?

brambl-sdk/src/test/scala/co/topl/brambl/validation/TransactionSyntaxInterpreterMintingCaseCSpec.scala
```diff
@@ -0,0 +1,565 @@
...
+import co.topl.brambl.models.{Datum, Event, TransactionOutputAddress}
+import co.topl.brambl.syntax._
+import scala.language.implicitConversions
+
+/**
+ * Test to coverage this specific syntax validation:
+ *  - Validations only for minting, After projection  (only if for all inputs and outputs isMint == true)
+ *    Case 3: Series
+ *  - asset minted correspond to token supply in series policy
+ *  -  the number of series constructor tokens burned matches the number of assets minted (only applies for limited token supply)
```
DiademShoukralla, line 17: Removing since this statement is not validated in these set of validations. This rule will be validated as part of Transfer-Series-Funds validation
```suggestion
```

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
nandotorterolo: Thanks for reviewing it, good catch. I will update using your logic, (with a small change because a.quantity >= tokenSupplied). 

>   (a.quantity: BigInt) % (tokenSupplied: BigInt) == 0 

a new test was created with your input suggestion.

mundacho (changes requested):

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,7 +393,29 @@ object TransactionSyntaxInterpreter {
...
+      transaction.inputs
+        .filter(_.value.value.isSeries)
+        .filter(sto =>
+          transaction.mintingStatements.map(_.seriesTokenUtxo).contains(sto.address) &&
+          a.seriesId.contains(sto.value.getSeries.seriesId)
+        )
+        .map(_.value.getSeries)
+        .headOption
+
+    val validAssets = assets.forall { a =>
```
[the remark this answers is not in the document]
nandotorterolo: let's chat about it, I think we covered the case for other assets, I will add a test to understand your point.
Would be great If I am wrong to create more unit tests on this case

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,25 +393,44 @@ object TransactionSyntaxInterpreter {
...
+
+              val sOut =
+                transaction.outputs
+                  .filter(_.value.value.isSeries)
+                  .filter(_.value.getSeries.seriesId == s.seriesId)
+                  .map(_.value.getSeries.quantity: BigInt)
+                  .sum
+
+              val burned = sIn - sOut
+              burned * tokenSupplied == (ams.quantity: BigInt)
```
DiademShoukralla, line 428: Per conversation this morning, `ams.quantity` should be replaced with a sum of all the Asset Minting Statements with the same seriesId

DiademShoukralla (approved):
LGTM!

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -438,7 +439,57 @@ object TransactionSyntaxInterpreter {
...
+              def quantity(s: Value.Series) = transaction.mintingStatements.map { ams =>
+                val filterSeries = transaction.inputs
+                  .filter(_.address == ams.seriesTokenUtxo)
+                  .filter(_.value.value.isSeries)
+                  .map(_.value.getSeries)
+                  .filter(_.seriesId == s.seriesId)
+                if (filterSeries.isEmpty) BigInt(0) else ams.quantity: BigInt
+              }
+
+              (ams.quantity: BigInt) <= s.quantity * tokenSupplied &&
```
mundacho, line 485: are we checking somewhere else that: `ams.quantity % tokenSupplied == 0`? If yes, then nothing else to add

mundacho (approved):
I think we are there, just double check that that validation is being made somewhere and we are there.

PR #117, opened 2023-09-25
BN-1219 - Update proposal immutable instances

nandotorterolo:
## Purpose
Immutable Instance for value: Update Proposal
## Approach
- implement immutable Instance, create a test for validation, (which actually does does not contains validation )

## Testing
preparePR

## Tickets
BN-1219


DiademShoukralla (approved):
LGTM

PR #119, opened 2023-09-26
TSDK-599 TransactionBuilder > Support Accumulator asset in new function buildTransferAllTransaction 

DiademShoukralla:
## Purpose

To support Assets with quantity descriptor=Accumulator

## Approach

- Removed old functions `buildGroupTransferTransaction`, `buildSeriesTransferTransaction`, `buildAssetTransferTransaction`, and `buildLvlTransferTransaction` and replaced with a combined `buildTransferAmountTransaction`. This is because it served no purpose having multiple of these functions since they did the same thing (transfer a certain amount of a certain token). For this function, assets of type ACCUMULATOR is still invalid.
- Created a new function `buildTransferAllTransaction` which serves to transfer ALL tokens (with the option to specify a specific type). For this function, we transferring ACCUMULATOR type is valid.
- For both functions `buildTransferAmountTransaction` and `buildTransferAllTransaction`, the only non-trivial parts are user input validation and building the UTXOs. 
    - They each have their own validations (`validateTransferAmountParams` and `validateTransferAllParams` respectively)
        -  The only change to `validateTransferAmountParams` is removing validation: all TXOs in the provided txos no longer need to be only LIQUID
        - `validateTransferAllParams` is new and verifies that all txos match the given LockPredicate, that there exist some TXOs that match a provided token identifier, and that there are enough LVLs to satisfy the fee.
    - building the utxos for both use the same function `buildUtxos`. Aggregating and calculating change within this function uses new `AggregationOps`. This trait provides interface for aggregating box values. A default implementation is provided where we can aggregate the quantities of all tokens except non-liquid asset types.
- Minor changes to Error handling; added `UserInputErrors` and removed `UnableToBuildTransaction`. This is facilitate debugging. The error itself is returned in these functions so we have access to the `cause` and stacktrace.
- updated existing tests to reflect these changes
- Added new tests for the new functionalities and supported behaviors: 
    - allow accumulator types in txos
    - added tests for AggregatorOps. Moved "typeIdentifier grouping" test case to this file to make use of `TransactionBuilderInterpreterSpecBase`
    - Added `TransactionBuilderInterpreterTransferAllSpec` for the new function `buildTransferAllTransaction`

## Testing

- ensured new and existing tests pass

## Tickets
* Closes TSDK-599


mundacho (approved):
LGTM!

PR #121, opened 2023-10-03
BN-1230 series transfer validation

nandotorterolo:
## Purpose

SeriesEqualFundsValidation
   *  - Check Moving Series Tokens: Let s be a series identifier, then the number of Series Constructor Tokens with group identifier s in the input is equal to the number of the number of Series Constructor Tokens with identifier s in the output.
   *  - Check Minting Constructor Tokens: Let s be a series identifier and p the series policy whose digest is equal to s, all of the following statements are true:
   *   -  The policy p is attached to the transaction.
   *   - The number of series constructor tokens with identifiers in the output of the transaction is strictly bigger than 0.
   *   -  The registration UTXO referenced in p is present in the inputs and contains LVLs.

## Testing
fix unit test

## Tickets
BN 1230


DiademShoukralla (approved):
LGTM!

mundacho (approved):
Approved!

PR #122, opened 2023-10-03
Test cleanup

DiademShoukralla:
## Purpose

cleaning up the tests. Most notably, the Mock Helpers and the builders package

## Approach

No changes to the logic of the tests. But removed a lot of unnecessary code by using more shared mock values. => Made the existing mockValues more composable 

## Testing

ensured all the tests still pass

## Tickets
* N/a, but doing before TSDK-598


mundacho (approved):
LGTM

PR #126, opened 2023-10-04
Bn 1204 validations only for minting case 3 step 3 alternative 2 DO not merge

nandotorterolo:
## Purpose

Pr to discuss alternative 2 of asset minting validation:
Specific discussion:
- TransactionSyntaxInterpreterMintingCaseCSpec
- test("Invalid data-input case 5, minting 2 Asset Token Limited, with 2 groups and 2 series") 

image

Note: this Pr replaces : https://github.com/Topl/BramblSc/pull/116

## Tickets
*

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
@@ -408,7 +408,10 @@ object TransactionSyntaxInterpreter {
       seriesIdOnMintedStatements(a).exists(s =>
         s.tokenSupply match {
           case Some(tokenSupplied) =>
-            (s.quantity: BigInt) * (tokenSupplied: BigInt) % (a.quantity: BigInt) == 0
+            // Is minted quantity a multiple of token supply?
+            (a.quantity: BigInt) % (tokenSupplied: BigInt) == 0 &&
+            // Is there enough series supply to mint asset quantity?
+            (s.quantity: BigInt) * (tokenSupplied: BigInt) >= (a.quantity: BigInt)
           case None => true
         }
       )
--- a/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
@@ -393,28 +393,44 @@ object TransactionSyntaxInterpreter {
       }
     }
 
-    def seriesIdOnMintedStatements(a: Value.Asset): Option[Value.Series] =
-      transaction.inputs
-        .filter(_.value.value.isSeries)
-        .filter(sto =>
-          transaction.mintingStatements.map(_.seriesTokenUtxo).contains(sto.address) &&
-          a.seriesId.contains(sto.value.getSeries.seriesId)
-        )
-        .map(_.value.getSeries)
-        .headOption
+    /**
+     * Let sIn be the total number of series constructor tokens with identifier in the input,
+     * sOut the total number of series constructor tokens with identifier in the output,
+     * and burned the number of where the referenced series specifies a token supply, then we have:
+     * sIn - burned = sOut
+     */
+    val validAssets = transaction.mintingStatements.forall { ams =>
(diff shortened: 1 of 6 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,7 +393,81 @@ object TransactionSyntaxInterpreter {
...
+        )
+
+      val ams = transaction.mintingStatements.find { stm =>
+        (getGroupStoGivenStm(stm), getSeriesStoGivenStm(stm)) match {
+          case (Some(g), Some(s)) => stm.groupTokenUtxo == g.address && stm.seriesTokenUtxo == s.address
+          case _                  => false
+        }
+      }
+
+      val group = getGroupStoGivenStm(ams.get).map(_.value.getGroup)
```
mundacho, line 420: maybe add a comment here stating why are we sure that ams will allways be != None. I'm guessing it is because we are in the minting projection. If it can be false we can maybe do a match.

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,7 +393,81 @@ object TransactionSyntaxInterpreter {
...
+      val series = getSeriesStoGivenStm(ams.get).map(_.value.getSeries)
+
+      (ams, group, series) match {
+        case (Some(ams), Some(g), Some(s)) =>
+          val b1 = (asset.quantity: BigInt) == (ams.quantity: BigInt) &&
```
mundacho, line 425: I we are checking this we also need to add the input assets I think

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,7 +393,81 @@ object TransactionSyntaxInterpreter {
...
+            } else {
+              val burned = ams.quantity / s.tokenSupply.get: BigInt
+              val outQuantity = (s.quantity: BigInt) - burned
+
+              val sOut = transaction.outputs
+                .filter(_.value.value.isSeries)
+                .map(_.value.getSeries)
+                .filter(sOut =>
+                  sOut.seriesId == s.seriesId &&
+                  (sOut.quantity: BigInt) == outQuantity &&
```
mundacho, line 452: This should consider the possibility of a sum. We interested in the sum of tokens.

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -393,7 +393,81 @@ object TransactionSyntaxInterpreter {
...
+            asset.seriesId.get == s.seriesId
+
+          val b2 =
+            if (s.tokenSupply.isEmpty) {
+              val sOut = transaction.outputs
+                .filter(_.value.value.isSeries)
+                .map(_.value.getSeries)
+                .filter(sOut =>
+                  sOut.seriesId == s.seriesId &&
+                  (sOut.quantity: BigInt) == (s.quantity: BigInt) &&
```
mundacho, line 438: We need to consider the sum here, no the exact quantity of the tokens. At any moment we might aggregate or deagreggate assets.

mundacho (changes requested):
There are still some things to fix. The cases where we aggregate or deaggregate are not considered.

```
Invalid Tx Example Which would pass this validation
===================================================


Input
-----

UTXO1: Group A
UTXO2: Group B
UTXO3: Series C (tokenSupply=5, quantity=3)
UTXO4: Series C (tokenSupply=5, quantity=3)

Ouput
-----

1: Group A
2: Group B
3: Series C (tokenSupply=5, quantity=1)
4: Series C (tokenSUpply=5, quantity=1)  <------ ⚠️
5: Series C (tokenSUpply=5, quantity=1)  <------ ⚠️
6: Asset A:C (quantity=10) 
7: Asset B:C (quantity=5)
```

mundacho (changes requested):
See comment above (this was a mistake)

PR #125, opened 2023-10-04
TSDK-597 Support Immutable Assets in buildTransferAllTransaction

DiademShoukralla:
## Purpose

Ensure we support Fractionable Assets in `buildTransferAllTransaction`.

## Approach

Mirrors https://github.com/Topl/BramblSc/pull/124 but for Immutable assets.

Existing implementation already supports. This PR adds/updates tests to ensure this.

- update Mock Data to also include immutable assets (group and series, group only, and series only fungible)
- Updated existing tests to consider the new mock data
- added new tests for buildTransferAllTransaction when immutable assets are identified to transfer

## Testing

ensure all tests pass

## Tickets
* closes TSDK-598

PR #127, opened 2023-10-04
Test Cleanup part 2

DiademShoukralla:
## Purpose

To make transaction builder tests more readable and maintainable; give readers of the code more visibility/understanding the different test cases.  This specifically affects the tests for `buildTransferAmountTransaction` and `buildTransferAllTransaction`

No new tests were added. This is simply refactoring how some of the existing tests are written.

## Approach

As of https://github.com/Topl/BramblSc/pull/125 , there are 24 tests for `buildTransferAllTransaction` and 33 tests for `buildTransferAmountTransaction` where we test different combinations of parameters while calling the functions.

As is, it is very hard for someone who is unfamiliar with the code to understand what the different test cases are testing. Two main reasons for this
- it is unclear which parameter(s) is being changed across tests
- it is unclear how the outputs should differ across tests

To address the first issue, we create helpers to run the functions; case classes `BuildTransferAmountTransaction` and  `BuildTransferAllTransaction` to aide `buildTransferAmountTransaction` and `buildTransferAllTransaction` respectively. These helpers provide default values for the function calls, allowing us to need only specify the parameter of interest for the tests.

for example

```
    val testTx = buildTransferAmountTransaction
      .withTokenIdentifier(groupValue.value.typeIdentifier)
      .run
```
better demonstrates that we are interested in testing the tokenIdentifier parameter

To address the second issue, we introduce a shared mock value `mockChange` which contains all the input values aggregated together. We can use this as a base and explicitly indicate the expected differences. 

for example

```
    // to recipient
    buildRecipientUtxos(List(assetGroup))
    ++
    // change due to excess fee and transfer input
    buildChangeUtxos(List(lvlValue, assetGroup))
    ++
    // change values unaffected by recipient transfer and fee
    buildChangeUtxos(
      mockChange.filterNot(v => List(LvlType, assetGroup.value.typeIdentifier).contains(v.value.typeIdentifier))
    )
```

better demonstrates which outputs are of interest to us for a given test case.

## Testing

ensure all tests pass

## Tickets
* n/a but doing this after/tracking with TSDK-597

### DIFF
--- a/brambl-sdk/src/test/scala/co/topl/brambl/builders/TransactionBuilderInterpreterAssetTransferSpec.scala
+++ b/brambl-sdk/src/test/scala/co/topl/brambl/builders/TransactionBuilderInterpreterAssetTransferSpec.scala
@@ -7,7 +7,8 @@ import co.topl.brambl.syntax.{
   int128AsBigInt,
   ioTransactionAsTransactionSyntaxOps,
   valueToQuantitySyntaxOps,
-  valueToTypeIdentifierSyntaxOps
+  valueToTypeIdentifierSyntaxOps,
+  LvlType
 }
 
 class TransactionBuilderInterpreterAssetTransferSpec extends TransactionBuilderInterpreterSpecBase {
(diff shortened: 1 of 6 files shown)


mundacho (approved):
Nice refactoring, I really appreaciated the changes

PR #131, opened 2023-10-11
BN-1246-update-proposal-validation

nandotorterolo:
## Purpose
- BN-1246-update-proposal-validation 

## Approach
-  new validation rule for proposal updated, simple rules, all values should be positives

## Testing
- unit test

## Tickets
BN-1246

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
@@ -324,8 +324,8 @@ object TransactionSyntaxInterpreter {
    */
   private def groupEqualFundsValidation(transaction: IoTransaction): ValidatedNec[TransactionSyntaxError, Unit] = {
 
-    val groupsIn = transaction.inputs.filter(_.value.value.isGroup).map(_.value.getGroup)
-    val groupsOut = transaction.outputs.filter(_.value.value.isGroup).map(_.value.getGroup)
+    val groupsIn = transaction.inputs.flatMap(_.value.value.group)
+    val groupsOut = transaction.outputs.flatMap(_.value.value.group)
 
     val gIds =
       groupsIn.groupBy(_.groupId).keySet ++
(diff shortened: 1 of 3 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -569,4 +570,25 @@ object TransactionSyntaxInterpreter {
     )
 
   }
+
+  private def updateProposalValidation(transaction: IoTransaction) = {
+    val upsIn = transaction.inputs.filter(_.value.value.isUpdateProposal).map(_.value.getUpdateProposal)
```
SeanCheatham, line 575: You can probably skip the validation of the inputs since they would have been validated already by the previous transaction.

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
```diff
@@ -569,4 +570,25 @@ object TransactionSyntaxInterpreter {
...
+    val upsOut = transaction.outputs.filter(_.value.value.isUpdateProposal).map(_.value.getUpdateProposal)
```
SeanCheatham, line 576: I think you can shorten this to `transaction.outputs.flatMap(_.value.value.updateProposal)`

SeanCheatham (approved):

DiademShoukralla (approved):

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
nandotorterolo: thanks for the suggestion

brambl-sdk/src/main/scala/co/topl/brambl/validation/TransactionSyntaxInterpreter.scala
nandotorterolo: thanks for the suggestion

PR #137, opened 2023-10-18
TSDK-604 Added Scaladoc to the portal

DiademShoukralla:
## Purpose

Adding the SDK's scaladoc to the portal for public consumption.

## Approach

- using SBT's unidoc plugin to generate and unify the Scaladocs for the projects and store them in the documentation site's static directory
    - Using unidoc since it gives us the ability to specify the target directory for the generated docs.
    - Storing them in the documentation site's `static/scaladoc/current` directory; under `static` b/c ability to serve pure HTML from this directory and under `current` to give us the ability to version the scaladocs in the future if need be.
    - Note that crypto's scaladoc is not being generated due to it's use of macros. We will need to revisit this in the future (since it also impedes our ability to upgrade to Scala3, remove the macros?)
- Run the unidoc command as part of `preparePR` for easier testing.
- Updated the workflow that deploys the Github pages; The scaladoc for the current version will be generated automatically as part of the process. 
- Minor changes in brambl-sdk to remove compilation warnings
- Updated docusaurus site to serve the scaladoc via links. + other minor changes

## Testing

- Tested locally with `npm run start` that everything works as expected
- Unfortunately, linking to the static scala docs does not work when running `npm run build` and `npm run serve` locally. So to ensure it would work, I modified  the deploy_docs trigger to push on all branches. The [site works as expected](https://topl.github.io/BramblSc/docs/current/reference/scaladoc).

## Tickets
* Part of TSDK-604


mundacho (approved):
Tested locally and in worked!

PR #143, opened 2023-10-31
TSDK-619 Set Current State

mundacho:
## Purpose

The goal of this task is to implement the set current state functionality in the wallet state api.

## Approach

- Implemented the functionality
- Applied the new terminology

## Testing

All tests are passing.

## Tickets

* TSDK-619


DiademShoukralla (approved):
LGTM

PR #144, opened 2023-11-02
TSDK-610 Add Support for TOPLs and new unknown types (UpdateProposals) in TransactionBuilder

DiademShoukralla:
## Purpose

Transfer functions in TransactionBuilder failed when TOPLs or UpdateProposals were passed in as parameters. The SDK was updated to support such parameters.

## Approach

- For UpdateProposals (and any new types that may be added to the ecosystem in the future), added ValueTypeIdentifier "UnknownType". If an "UnknownType" is provided as a parameter (either in TXOs or as tokenIdentifier) in TxBuilder functions, a validation error message will be returned.
- Support Transferring TOPLs; transferAll supports all TOPLs whereas transferAmount supports all TOPLs in the txos but only TOPLs without staking registration for tokenIdentifier (since as a default we won't aggregate/deaggregate for the user).
- Added tests

## Testing

- Added new tests for ValueTypeIdentifier, Aggreations, and TransactionBuilder
- Ensured all tests pass

## Tickets
* Closes TSDK-610


mundacho (approved):
Nice! Approved!

PR #148, opened 2023-11-09
TSDK-632 Fix bug.

mundacho:
## Purpose

Fix bug that made tests flaky. 

## Approach

Verified that the bug was happening because the keys parts of the generated derived private keys need to be 32 bytes long. Sometimes, this was not the case because of the implementation of BigInt. After checking the [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) standard, we needed to implement the sec_256 serialization for big integers.

The implementation guaranteed that all numbers are always 32 bytes in length and thus the bug cannot happen again.

## Testing

A scala-cli script was developed to reproduce the bug:

```scala
// You can run this code using scala-cli. Save it in a file called `testBug.sc` and run it with `scala-cli testBug.sc`
//> using scala 2.13
//> using repository sonatype-s01:releases
//> using dep co.topl::service-kit:2.0.0-alpha7
//> using dep org.typelevel::cats-core:2.10.0

import cats.effect.kernel.Resource
import cats.effect.IO
import co.topl.brambl.wallet.WalletApi
import co.topl.brambl.servicekit.WalletStateApi
import co.topl.brambl.constants.NetworkConstants
import co.topl.brambl.servicekit.WalletKeyApi
import java.sql.{Connection, DriverManager}
import cats.effect.std
import io.circe.syntax._
import co.topl.crypto.encryption.VaultStore.Codecs._
import cats.effect.unsafe.implicits.global

case class CreateWallet(
    file: String,
    password: String
) {
  val walletKeyApi = WalletKeyApi.make[IO]()
  val walletApi = WalletApi.make(walletKeyApi)

  val createWallet = (for {
    wallet <- walletApi
      .createNewWallet(
        password.getBytes(),
        Some("passphrase")
      )
      .map(_.fold(throw _, identity))
    keyPair <- walletApi
      .extractMainKey(
        wallet.mainKeyVaultStore,
        password.getBytes()
      )
      .flatMap(
        _.fold(
          _ =>
            IO.raiseError(
              new Throwable("No input file (should not happen)")
            ),
          IO(_)
        )
      )
    derivedKey <- walletApi.deriveChildKeysPartial(keyPair, 1, 1)
  } yield ())//.handleErrorWith(_ => IO.trace)

}

val file = "myWallet"
val password = "password"

import cats.syntax.parallel._

// run using:
for (i <- 1 to 100)
  yield CreateWallet(file + i + ".db", password).createWallet.unsafeRunSync()
```

It allowed to reproduce the bug and find a solution. After the fix, the bug does not appear anymore.

## Tickets
* TSDK-625

### DIFF
--- a/crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
+++ b/crypto/src/main/scala/co/topl/crypto/signing/ExtendedEd25519.scala
@@ -127,19 +127,20 @@ class ExtendedEd25519 extends EllipticCurveSignatureScheme[SecretKey, PublicKey]
     val zRight =
       BigInt(1, z.slice(32, 64).reverse)
 
-    val nextLeft =
-      ByteBuffer
-        .wrap((zLeft * 8 + lNum).toByteArray.reverse)
-        .order(ByteOrder.LITTLE_ENDIAN)
-        .array()
-        .take(32)
+    // serialize a 32 byte BigInt to a 32 byte array according to the standard
+    // in https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
+    // (see standard conversion functions)
+    // example Java implementation at:
+    // https://github.com/bloxbean/cardano-client-lib/blob/00f69a0c909770b919c4cc4834898cfd099f38e9/crypto/src/main/java/com/bloxbean/cardano/client/crypto/bip32/util/BytesUtil.java#L61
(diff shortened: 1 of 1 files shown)


DiademShoukralla (approved):
I ran the script on my machine multiple times (for a total of 1000 calls to `CreateWallet`) and they all passed

PR #150, opened 2023-11-10
TSDK-631 List interactions support

mundacho:
## Purpose

The goal of this PR is to add support to list interactions to the brambl sdk/service-kit.

## Approach

We implemented new function in the `WalletStateAlgebra`.

## Testing

Two unit tests were added.

## Tickets
* TSDK-631


DiademShoukralla (approved):
LGTM!

PR #151, opened 2023-11-11
TSDK-634 Document Prove Transaction  (+added examples for existing docs)

DiademShoukralla:
## Purpose

Added documentation and example for how a user can prove a  transaction

## Approach

This involved some minor changes to code
- Added object with trait `WalletStateResource` so users can easily obtain a wallet state resource
- Added trait and object for `RpcChannelResource` so users can easily obtain a grpc channel resource.  The implementation was ported over from brambl-cli. This also involved adding extra library dependencies to this repo

Added example to RPC `queryUtxos` and transfer transactions. This is because these code-snippets were being used in the `prove` example so it was a matter of copy-pasting

Added documentation and example for `prove`

This PR does not include documentation for `proveAndValidate` (will be added separately) as there is some discussion needed around how the user should obtain a Quivr Context to validate with.

## Testing

Tested locally

## Tickets
* Part of TSDK-626
* Closes TSDK-634


mundacho (approved):
LGTM! Just one question. was it really necessary to include the scalapb plugin?

DiademShoukralla:
@mundacho Thanks for the review and pointing that out. I tested without it and it works; only `grpc-netty` is required

PR #155, opened 2023-11-15
TSDK-637 Workflow dependencies

DiademShoukralla:
## Purpose and Approach

- readded `+test` to `preparePR`. I verified that it works on my machine when using WSL
- updated GH Actions workflows such that maven release and GH Pages do not run unless build/test succeeds. Update to documentation portal will only occur if we are pushing to main
- Added Publish Test Results
- Added GH Deployments for releases (tag pushes) and snapshots (branch pushes)

## Testing

For push to any branch:
https://github.com/Topl/BramblSc/actions/runs/6893972552

For push to main; I will manually check that the workflow ran properly once merged. Should be similar to any branch

For tags; I will manually check that the workflow ran properly once a release is created. Should deploy pages and deploy to releases 

## Tickets
* closes TSDK-637

### DIFF
--- a/.github/workflows/_sbt_build.yml
+++ b/.github/workflows/_sbt_build.yml
@@ -78,7 +78,12 @@ jobs:
         with:
           name: Test Results (Java ${{ matrix.java }})
           path: target/test-reports/*.xml
-      
+      - name: Publish Test Results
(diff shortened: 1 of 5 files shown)


mundacho (approved):
Looks great!

PR #156, opened 2023-11-17
TSDK-638 Filter out UnknownType

DiademShoukralla:
## Purpose

If the input parameter `txos` (in TransactionBuilder functions) contain an UnknownType parameter, filter it out instead of sending an error.

The reason for this change is that when spending from Genesis, there will always be an UpdateProposal value, thus always erroring out (forcing users to manually filter them out before hand). This became apparent when writing examples in the Documentation portal; for example

```
txos <- genusQuery.queryUtxo(fromAddr)
fromTxos = txos.filter(_.transactionOutput.value.value.typeIdentifier != UnknownType) // <-- this line
res <- transactionBuilderApi.buildTransferAllTransaction(fromTxos, fromLock.getPredicate, toAddr, toAddr, 1L)
```

## Approach

- Allowed UnknownType to be provided and filter it out within the function (instead of expecting the user to do it before hand)
- updated tests
- Updated documentation examples

## Testing

- Ensured all unit tests pass
- Ensured all the updated examples can successfully be run

## Tickets
* closes TSDK-638

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/builders/TransactionBuilderApi.scala
@@ -31,6 +31,7 @@ import co.topl.brambl.syntax.{
   valueToQuantitySyntaxOps,
   valueToTypeIdentifierSyntaxOps,
   LvlType,
+  UnknownType,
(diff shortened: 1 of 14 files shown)


mundacho (approved):
LGTM!

PR #161, opened 2023-11-24
TSDK-641 Usability Improvements for CLI (get block by depth)

mundacho:
## Purpose

As part of improving the usability of the CLI, we implemented full validation at broadcast time. This  validation requires to know the current height of the block and the current tick. Although this functionality is implemented in the node, the brambl-sdk did not support it. We implemented it in order to add the validation to the CLI.

## Approach

We implemented it by following the same pattern as before.

## Testing

All tests continue to pass.

## Tickets
* TSDK-641


DiademShoukralla (approved):
LGTM

PR #166, opened 2023-12-12
TSDK-666 initWalletState parameter improvement

DiademShoukralla:
## Purpose

Ease of use improvement for using `initWalletState`.  Now users do not have to do a (x=1,y=1) partial derivative of their Topl Main key pair before initializing a wallet state.

## Approach

- Updated Signature of the function to take in a KeyPair (needed for partial derivation).
- Updated Service Kit to perform the partial derivation within the `initWalletState` function
- Updated all callers (tests) to use the updated parameter
- Updated documentation to use the updated parameter

## Testing

- Ensure all tests pass
- Ensure all documentation examples run successfully. Note: The code examples within "Service Kit > Usage" are not directly runnable using `scala-cli` *yet*. This is because it depends on a released version of the SDK. These changes are not yet released but will be a part of `beta2` (which is what the dependency references). In other words, this *will* be runnable once `beta2` is released (which is when the documentation portal will be updated). Despite this, I ensured running the code in a worksheet works as expected (same as the other code examples).
- Used in `brambl-cli` and ensured the unit tests and integration tests all pass.

## Tickets
* TSDK-666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added a tutorial for creating a new custom asset
 
+### Changed
+
+- Service Kit's initWalletState now expects a Topl main key pair instead of a (1,1) partial derivative verification key
+
 ## [v2.0.0-beta1] - 2023-12-05
 
 ### Added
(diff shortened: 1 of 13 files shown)


mundacho (approved):
LGTM!

PR #172, opened 2024-01-29
TSDK-729 Support for Sha256 Digest Locks

DiademShoukralla:
## Purpose

With the upcoming BTC bridge, we need to support Sha256 Digest propositions in our ecosystem (so the preimages and digests are cross compatible with bitcoin). This PR adds this support.

## Approach

- Propositions themselves are routine-agnostic, so no changes were needed to the quivr package
- We already have a Sha256 implementation, so no changes were needed to the crypto package
- In brambl-sdk, added a Sha256DigestInterpreter which validates a given Digest and Preimage using the Sha256 routine.
- In brambl-sdk, added the "Sha256" routine to `hashingRoutines` in the Context object; this routine uses the newly created Sha256DigestInterpreter, allowing the quivr verifier to validate for any Digest propositions using a "Sha256" routine.
-  In the docs, we now list the possible values for `routine` (for both digest and signature operations) 

## Testing

- Ensured exiting tests pass
- Added a test for Proposition Templates (sanity): Ensure a Digest Template with routine "Sha256" creates a Digest Proposition with routine "Sha256"
- Added a test for proving and verifying: Ensure that a Transaction requiring both Blake2b256 and Sha256 digest propositions to be satisfied can be proven and valid, given that the WalletStateApi contains the relevant preimages. This tests that the "Sha256" routine within Context works as expected.

## Tickets
* Closes TSDK-729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - YYYY-MM-DD - TODO replace date after release 
 
+### Added
+
+- Support for Sha256 Digest Propositions in the SDK. This change allows successful Quivr validation of a Digest Proposition with routine="Sha256".
+
 ## [v2.0.0-beta2] - 2024-01-10
 
 ### Added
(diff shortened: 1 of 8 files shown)


mundacho (approved):
Looks OK, just one question: I noticed that in Bitcoin the secret needs to be 32 bytes long, or at least, that's what the miniscript compiler says :/. Do we need such a restriction on our side? Did you include that?

DiademShoukralla:
Per discussion, we will leave this as-is, and provide the secret generation infrastructure to handle to 32byte restriction

PR #177, opened 2024-03-12
TSDK-757 Streams for Bitcoin Block Monitoring

DiademShoukralla:
## Purpose

Added support for monitoring Bitcoin blocks via fs2 streams

## Approach

- Implemented a monitor that takes a reference to a bitcoind instance (either local or remote) and reports on any new block
- We also report on any block that has occurred since an existing block hash
- This is returned to the user as an infinite fs2 stream

## Testing

- Manually tested
- Tests are implemented in a separate PR https://github.com/Topl/BramblSc/pull/178

## Tickets
* Tracked under a closed ticket TSDK-757
* Relates to TSDK-766


mundacho (approved):
LGTM!

PR #183, opened 2024-04-26
TSDK-758 Monitor Bifrost Blocks (live)

DiademShoukralla:
## Purpose

Similar to how we monitor incoming Bitcoin blocks, we should monitor incoming Bifrost blocks.

## Approach

- Added `SynchronizationTraversal` to the DataApi bifrost query algebra
- Added a BifrostMonitor class and object for which incoming blocks will be returned live via a stream
- The blocks returned can either represent an Applied or Unapplied block. Both of which would contain a stream of the contained transactions

## Testing

Added 2 integration tests

- Verify that multiple live incoming applied blocks will be reported as expected
- Verify that the transactions within an incoming applied block is reported as expected

## Tickets
* closes TSDK-758

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BifrostMonitor.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BifrostMonitor.scala
@@ -17,14 +17,18 @@ import fs2.Stream
  * @param startingBlocks Past blocks that should be reported.
  */
 class BifrostMonitor(
-  blockIterator:     Iterator[SynchronizationTraversalRes],
-  getFullBlock: BlockId => IO[FullBlockBody],
-  startingBlocks: Vector[BifrostBlockSync],
+  blockIterator:  Iterator[SynchronizationTraversalRes],
+  getFullBlock:   BlockId => IO[FullBlockBody],
+  startingBlocks: Vector[BifrostBlockSync]
 ) {
-  def pipe(in: Stream[IO, SynchronizationTraversalRes]): Stream[IO, BifrostBlockSync] = in.evalMap(sync => sync.status match {
-    case Applied(blockId) => getFullBlock(blockId).map(AppliedBifrostBlock)
-    case Unapplied(blockId) => getFullBlock(blockId).map(UnappliedBifrostBlock)
-  })
+
+  def pipe(in: Stream[IO, SynchronizationTraversalRes]): Stream[IO, BifrostBlockSync] = in.evalMap(sync =>
+    sync.status match {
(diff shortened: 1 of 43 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BifrostMonitor.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Initialize and return a BifrostMonitor instance.
+   * @param bifrostQuery The bifrost query api to retrieve updates from
+   * @param startBlock The blockId of a past block. Used to retroactively report blocks. The bifrost monitor will report all blocks starting at this block.
+   * @return An instance of a BifrostMonitor
+   */
+  def apply(
+    bifrostQuery: BifrostQueryAlgebra[IO],
+    startBlock:   Option[BlockId] = None
```
mundacho, line 65: This parameter `startBlock` is never used.

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BifrostMonitor.scala
```diff
@@ -0,0 +1,78 @@
...
+  ): IO[BifrostMonitor] = {
+    def getFullBlock(blockId: BlockId): IO[FullBlockBody] = for {
+      block <- bifrostQuery.blockById(blockId)
+    } yield block match {
+      case None                 => throw new Exception(s"Unable to query block ${display(blockId)}")
+      case Some((_, _, _, txs)) => FullBlockBody(txs)
+    }
+    for {
+      updates <- bifrostQuery.synchronizationTraversal()
+    } yield new BifrostMonitor(updates, getFullBlock, Vector.empty)
```
mundacho, line 75: Is this meant to be hardcoded?

mundacho (approved):
LGTM, just two small comments. Since I found a non used variable I enabled that check at the compiler level. There were several unused imports. I fixed those with the IDE and commited the changes.

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BifrostMonitor.scala
DiademShoukralla: That was intentional, it is a placeholder and will be utilized for the current ticket I am working on

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BifrostMonitor.scala
DiademShoukralla: If you are referring to Vector.empty, similar to the other comment, it will replaced in the current ticket I am working on

PR #187, opened 2024-05-10
Hot fix: Fix responsiveness warning

DiademShoukralla:
## Purpose

Address warning that appears when using the Bifrost monitor

```
[WARNING] Your app's responsiveness to a new asynchronous event (such as a new connection, an upstream response, or a timer) was in excess of 100 milliseconds. Your CPU is probably starving. Consider increasing the granularity of your delays or adding more cedes. This may also be a sign that you are unintentionally running blocking I/O operations (such as File or InetAddress) without the blocking combinator.
```

## Approach

Replace `Stream.fromIterator` with `Stream.fromBlockingIterator`

## Testing
(manual)
Ran the bifrost monitor for a long period of time, previously , the warning would appear. I have not encountered the warning since

## Tickets
* n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - YYYY-MM-DD - TODO replace date after release 
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


mundacho (approved):
Genius! LGTM!

PR #188, opened 2024-05-16
Bitcoin monitor >  add height field

DiademShoukralla:
## Purpose

The blocks within the bitcoin monitor stream did not contain block-height information. This PR adds this information

## Approach

- For both retroactive and live block updates, query the bitcoin RPC  to retrieve block information (via getBlockHeight) and add it to the BitcoinBlock case class
- The live reporting required giving Zmq (addToQueue and initZmqSubscriber) access to the BitcoindRpcClient instance.

## Testing

In the existing tests for BitcoinMonitor, ensure the height fields are what we expect

## Tickets
* n/a


mundacho (approved):
Excellent!

PR #189, opened 2024-05-21
TSDK-801 Support Bitcoin Reorgs to the Bitcoin Monitor

DiademShoukralla:
## Purpose

To capture affected blocks when a bitcoin reorg occurs.

## Approach

If a reorg is detected (incoming block's parent does not match the last reported block), we find the common ancestor of the two. From this common ancestor, we unapply the chain (emit UnapplyBitcoinBlocks) leading to the previous last reported block. We also apply the new chain (emit AppliedBitcoinBlocks) leading to the newly adopted block.

## Testing

Added an integration test

## Tickets
* Closes TSDK-801

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
@@ -3,8 +3,8 @@ package co.topl.brambl.monitoring
 import akka.actor.ActorSystem
 import cats.effect.IO
 import cats.effect.std.Queue
-import co.topl.brambl.monitoring.BitcoinMonitor.BitcoinBlockSync
-import fs2.Stream
+import co.topl.brambl.monitoring.BitcoinMonitor.{AppliedBitcoinBlock, BitcoinBlockSync, UnappliedBitcoinBlock}
+import fs2.{Pipe, Stream}
 import org.bitcoins.core.config.NetworkParameters
 import org.bitcoins.core.protocol.blockchain.Block
 import org.bitcoins.core.protocol.transaction.Transaction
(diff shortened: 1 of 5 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
```diff
@@ -32,19 +32,96 @@ import scala.concurrent.{Await, Future}
...
+  blockQueue:        Queue[IO, AppliedBitcoinBlock],
+  startingBlocks:    Vector[AppliedBitcoinBlock],
+  initZmqSubscriber: Queue[IO, AppliedBitcoinBlock] => ZMQSubscriber,
+  bitcoind:          BitcoindRpcClient
 ) {
   val subscriber: ZMQSubscriber = initZmqSubscriber(blockQueue)
   subscriber.start()
 
+  // The last live block that was reported by the monitor. Used to detect reorgs
+  var currentTip: Option[AppliedBitcoinBlock] = startingBlocks.lastOption
```
mundacho, line 44: I think this is unsafe, the safe alternative is to use `Ref[IO, Option[AppliedBitcoinBlock]]`. You can find the documentation here: https://typelevel.org/cats-effect/docs/std/ref. I use it in the bridge if you want to see some example. Sadly, this will require some modifications of the way you build you object (you probably need to pass that reference in the constructor).

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
```diff
@@ -32,19 +32,96 @@ import scala.concurrent.{Await, Future}
...
+   */
+  private def synchronizeChain: Pipe[IO, AppliedBitcoinBlock, Vector[BitcoinBlockSync]] = in =>
+    in.evalMap { adoptedBlock =>
+      // Checking if the new block's parent matches our last reported block
+      if (
+        currentTip.isEmpty || currentTip
+          .map(_.block.blockHeader.hashBE)
+          .contains(adoptedBlock.block.blockHeader.previousBlockHashBE)
+      ) {
+        currentTip = Some(adoptedBlock)
```
mundacho, line 68: For example here, using the `Ref` you will be able to have a clean return value that is at the same time safe.

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
```diff
@@ -32,19 +32,96 @@ import scala.concurrent.{Await, Future}
...
+          newBlockHeight <- IO.fromFuture(
+            IO(bitcoind.getBlockHeight(newTip.head.block.blockHeader.previousBlockHashBE))
+          )
+        } yield AppliedBitcoinBlock(newBlock, newBlockHeight.get) +: newTip // height should be 1 less than before
+        else IO.pure(newTip)
+      updatedOldTip <-
+        if (
+          newTip.head.height < oldTip.head.height || (newTip.head.height == oldTip.head.height && newTip.head.block.blockHeader.hashBE == newTip.head.block.blockHeader.hashBE)
+        ) for {
+          oldBlock <- IO.fromFuture(IO(bitcoind.getBlockRaw(oldTip.head.block.blockHeader.previousBlockHashBE)))
```
mundacho, line 109: Just to be clear. This assumes that bitcoin keeps the blocks that were there before the reorg, is this correct? Any sources maybe to document the assumption?

mundacho (approved):
Looks good to me! There are just 2 small comments to address.

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
DiademShoukralla: Yes, you are correct about my assumption. This assumption only holds if the bitcoin node does **not** have pruning enabled. Pruning is disabled by default. 
Read more here: https://bitcoincore.org/en/releases/0.11.0/#block-file-pruning

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
DiademShoukralla: Thanks for the review! I will implement this

PR #191, opened 2024-05-24
TSDK-811 Added Height Field to Bifrost Monitor

DiademShoukralla:
## Purpose

Add a height field in the Bifrost monitor stream

## Approach

We already queried bifrost for the Transactions. We utilized this call to retrieve the height from the header as well

## Testing

- Verified accuracy of returned heights in the existing tests
- These tests found a bug (now fixed) that there was a gap in between reporting retroactive blocks and live blocks. The reason is because `.blockByDepth(1)` did not return the tip of the chain. This was replaced with `blockByDepth(0)`.

## Tickets
* closes TSDK-811

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,4 @@ docker/*.jar
 
 # Scala Worksheet files
 *.sc
+playground/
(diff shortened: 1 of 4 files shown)


mundacho (approved):
LGTM!

PR #193, opened 2024-07-08
TSDK-852 Changes to BifrostMonitor and BitcoinMonitor

DiademShoukralla:
## Changes

- Added integration test for the Bifrost Monitor to ensure reorgs are handled as expected
- Added the bifrost RPC function "makeBlock" to the BifrostQuery API. This is to force block production and is only meant to be used in "RegTest" mode.
- Bifrost and Bitcoin monitor APIs now return a resource containing their respective block stream.
- Retroactive reporting of existing blocks in the bifrost monitor is now inclusive (includes the starting block).
- Bifrost monitor's update stream is now self-contained. It's channel will now shutdown when the monitor is released. 
- The ZMQ subscriber that is used in the Bitcoin Monitor now stops when the monitor is released. Users no longer have to explicitly call ".stop()"
- Updated existing BitcoinMonitor and BifrostMonitor tests (including the workflow in GH actions) to account for these changes.

## Testing

- Updated existing integration tests
- Added new integration test for BifrostMonitor reorgs
- Ensured these tests ran successfully locally and in github actions

## Tickets
* TSDK-852

### DIFF
--- a/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
@@ -249,7 +249,6 @@ object BitcoinMonitor {
     }
 
     val existingHashes = getBlockHashes(startBlock)
-    println("Retroactively fetching blocks:")
     existingHashes.foreach(h => println(h.hex))
     (for {
       blockQueue <- Queue.unbounded[IO, AppliedBitcoinBlock]
--- a/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
+++ b/brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
@@ -35,11 +35,11 @@ import scala.concurrent.{Await, Future}
  * @param currentTip Reference to the most recent chain tip reported by the monitor
  */
 class BitcoinMonitor(
-  blockQueue:        Queue[IO, AppliedBitcoinBlock],
-  startingBlocks:    Vector[AppliedBitcoinBlock],
-  bitcoind:          BitcoindRpcClient,
-  currentTip:        Ref[IO, Option[AppliedBitcoinBlock]],
-  val subscriber: ZMQSubscriber,
+  blockQueue:     Queue[IO, AppliedBitcoinBlock],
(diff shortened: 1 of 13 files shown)


brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
```diff
@@ -248,30 +243,27 @@ object BitcoinMonitor {
...
-      startingBlocks <- IO.fromFuture(
-        IO(
-          Future.sequence(
-            existingHashes.map(hash =>
-              for {
-                b <- bitcoind.getBlockRaw(hash)
-                h <- bitcoind.getBlockHeight(hash)
-              } yield AppliedBitcoinBlock(b, h.get)
-            ) // .get won't cause issue since the getBlockHeight hardcodes it as Some(_)
+    Resource
```
mundacho, line 246: This is how I would do it, for more clarity and safety. You can remove the public subscriber reference from the monitor. In this way, after the monitor is removed, the subscriber will automatically be closed.
```
for {
blockQueue <- Queue.unbounded[IO, AppliedBitcoinBlock].toResource
startingBlocks <- IO.fromFuture(
  IO(
    Future.sequence(
      existingHashes.map(hash =>
        for {
          b <- bitcoind.getBlockRaw(hash)
          h <- bitcoind.getBlockHeight(hash)
        } yield AppliedBitcoinBlock(b, h.get)
      ) // .get won't cause issue since the getBlockHeight hardcodes it as Some(_)
    )
  )
).toResource
currentTip <- Ref.of[IO, Option[AppliedBitcoinBlock]](startingBlocks.lastOption).toResource
subscriber <- Resource.make(IO(initZmqSubscriber(bitcoind, zmqHost, zmqPort)(blockQueue)))(_.stop())
} yield new BitcoinMonitor(blockQueue, startingBlocks, bitcoind, currentTip, subscriber)
```

mundacho (approved):
LGTM, please consider my comment. If you have questions you can contact me.

brambl-sdk/src/main/scala/co/topl/brambl/monitoring/BitcoinMonitor.scala
DiademShoukralla: Thanks for the feedback and example!

PR #196, opened 2024-08-01
Fix maven deploy

DiademShoukralla:
## Purpose

The maven-release job keeps failing due to Too many Server Redirects (20). We are testing out a possible fix done by normalizing the artifact names (removing "+" from the version).

## Approach

Applied the fix done in https://github.com/Topl/topl-btc-bridge/commit/ceb35cfb3b0c4c0063c57e5511d06ac152436cf1

## Testing

Testing is done manually by seeing if the maven-release job is successful 

## Tickets
n/a

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,20 @@ lazy val commonSettings = Seq(
   testFrameworks += TestFrameworks.MUnit
 )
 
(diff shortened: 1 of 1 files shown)


mundacho (approved):
LGTM!

PR #198, opened 2024-08-14
TSDK-711 Merging Tutorial

DiademShoukralla:
## Purpose

To write documentation and a tutorial for how to merge assets

## Approach

- Added Reference page for the new transaction builder function
- Added tutorial

**Note:** The tutorial does not broadcast the final merging transaction. This is because node-side validation is not yet implemented. Instead we display the built transaction

## Testing

- Locally ran docusaurus and ensured all changes show up as expected
- Locally ran each of the code examples and ensured they return the output as expected

## Tickets
* closes TSDK-711


mundacho (approved):
LGTM!

